PR #14, opened 2020-08-03
Add documentation on index dates

evansd:
These were added in this PR here: https://github.com/opensafely/cohort-extractor/pull/255

The main reason for adding them is to support extracting data for measures in the aftershocks project. It means that the same study definition can be run with the index date set to successive months in order to generate monthly timeseries data.

### DIFF
--- a/study_definition.md
+++ b/study_definition.md
@@ -497,6 +497,47 @@ There is also a custom distribution matching a typical population-age distributi
         },
     ),
 ```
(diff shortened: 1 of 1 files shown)


evansd:
You can't at the moment because I didn't want to implement a full parser for date expressions, but we could certainly extend it to support those if it turned out to be useful.

PR #18, opened 2020-08-12
Document new codelist pages

ghickman:
This updates the section for adding new Codelists, removing the admin references, and adds a section to explain updating an existing Codelist.

Fixes opensafely/opencodelists#52

### DIFF
--- a/codelist creation.md
+++ b/codelist creation.md
@@ -94,3 +94,10 @@ How contributions to codelists are acknowledged -- to be agreed.
 	* This will show you the latest version for a Codelist
 * If it's a draft version there will be a Publish version button on the left below Create new version.
 * If not, it's already published, good job!
+
+
+## Adding a Codelist Version
+
+* Go to an existing Codelist page.
+* Click Create new version
+* If you want to update the existing version, click Update version instead.
--- a/codelist creation.md
+++ b/codelist creation.md
@@ -80,4 +80,9 @@ How contributions to codelists are acknowledged -- to be agreed.
 
 ## Editing existing codelists
 
-CUrrently undocumented.
+* Go to an existing Codelist page.
+* Click Edit metadata.
+* Edit the relevant fields
+	* Note: Changing the CSV data requires you Update the current Version or Create a new Version, both can be done from the Codelist page.
+	* Add, remove, or edit the References and SignOffs as needed.
+* Click Submit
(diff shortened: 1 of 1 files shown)


codelist creation.md
```diff
@@ -60,13 +60,11 @@ Once a draft codelist has been agreed, it must be signed-off by a "data expert"
 
 ## Add to [codelists.opensafely.org](https://codelists.opensafely.org)
 
-* Go to the openSAFELY [codelist admin page](https://codelists.opensafely.org/admin). You will need an editor account.
-* Under "CODELISTS", click "Add".
+* Go to the openSAFELY [new codelist page](https://codelists.opensafely.org/codelist/opensafely/). You will need an editor account.
```
inglesp, line 63: "Ask one of the tech team for one if you do not have one."

codelist creation.md
```diff
@@ -60,13 +60,11 @@ Once a draft codelist has been agreed, it must be signed-off by a "data expert"
...
 * Fill in the fields. Include lots of detail (specific guidance to follow).
-	* **Version**: this should be the current date.
 	* **CSV data**: _DO NOT_ copy the data from within spreadsheet software such as Excel (this can reformat or revalue cells, for example with very long numeric codes that are interpreted as integers). Instead, open the CSV file in a text editor such as notepad and copy the raw text data.
```
inglesp, line 65: We're no longer copying/pasting CSV data into a `textarea`, so this bit is out of date.  Instead, could you add a line about how to create a CSV file by exporting a spreadsheet from Excel?  I know that @brianmackenna has done this recently.

brianmackenna: So in excel there are two ways to change the file type from `xlsx` to `csv`. 

I recommend if you have newer versions of excel(see screenshot below)
1. Go to file
2. Go to "export" on the left handside ribbon
3. Select "change file type"

The second way is to
1. File
2. Save as
3. change the file type to .`csv`

This is the number way to do it on google but I have found this buggy. It makes a file that looks like a csv but when we have uploaded it to opencodelists it has broken. When its been investigated it turns out is not quite a csv. I have found the first way to work ok so far. 

image

codelist creation.md
ghickman: Fixed in a944b80

codelist creation.md
ghickman: Fixed in 155a981

PR #19, opened 2020-08-13
SNOMED Generated from CTV3

ghickman:
This documents that most of our SNOMED CT codelists are generated from CTV3 codelists and how that's done.

@inglesp – do you think anything else needs to be said here?


snomed.md
```diff
@@ -0,0 +1,3 @@
+# SNOMED CT Codelists
+
+At the time of writing (2020-08-13) most SNOMED CT Codelists have been generated from CTV3 Codelists using a [mapper function](https://github.com/opensafely/opencodelists/blob/master/mappings/ctv3sctmap2/mappers.py).
```
inglesp, line 3: Could you add

> They have not been reviewed, or used in any research.

ghickman: Fixed in dbd9d67

PR #22, opened 2020-08-25
add explanation for missing values

wjchulme:


study_definition.md
```diff
@@ -425,6 +425,9 @@ household_size=patients.household_as_of(
 ),
 ```
 
+## Missing values
+
+If a query returns no matching record for a patient &mdash; for example if there are no blood pressure values recorded in a given period, or if there is no death date because the patient hasn't died, or if there is no household size available &mdash; then a default value will be returned. For strings and dates, the default value is the empty string `""`. For booleans, integers, or floats, the default value is `0`. There is no universal `null` value outputted to `input.csv`.
```
sebbacon, line 430: ...and this is because handling nulls in CSVs is prone to all sorts of horrible errors (e.g. in pandas it assumes NA is a null value by default, which wreaks havoc when doing analysis of sodium pathology results....

(not sure if this needs explaining in the docs or not, I'll leave that to you!

sebbacon (approved):

study_definition.md
wjchulme: Oh good. I wasn't sure the "no null values" thing was true. I've added a small explanation

wjchulme:
Just a note that this default behaviour might cause issues down the line if we're retrieving a variable for which `0` is a valid value.

sebbacon:
Yes, I have a vague memory of discussing this with @evansd and concluding there were currently no such cases we could think of, or at least none where we would need to distinguish between 0 and null.

PR #60, opened 2020-12-07
Will bits

wjchulme:


docs/pipelines.md
```diff
@@ -116,75 +117,87 @@ cohortextractor run dummy generate_cohorts expectations
...
 
 You will also find a new folder, `metadata/`, which contains logging information about your run. If any of your actions fail, you may find clues here as to why.
 
+Each action is run in its own isolated environment in a temporary working directory. This means that all the necessary libraries and data must be imported within the script for each action &mdash; For R users, this essentially means that the R session is restarted for each action. 
+
+### Dependencies
+
+If one or more dependencies of an action have not been run (i.e., their outputs do not exist) then these dependency actions will be run first. If a dependency has changed but has not been run (so the outputs are not up-to-date with the changes), then the dependent actions will be run on the old outputs. To force the dependencies to be run, use the `--force-run-dependencies` option.
+
+For example `cohortextractor run run_model --force-dependencies` will ensure that both the `run_model` and `generate_cohorts` actions are run, even if `input.csv` already exists.
```
evansd, line 154: This command isn't quite right. If you're doing it through the `cohortextractor` then you need the fairly hideous command:
```
cohortextractor run dummy run_model expectations --force-run --force-run-dependencies
```

The new command will accept:
```
opensafely run run_model --force-run-dependencies
```

Or the even shorter:
```
opensafely run run_model -f
```

evansd (approved):
:+1: generally looks good. There's a mistake with the `force-run-dependencies` command but that will all be simplified soon anyway.

docs/pipelines.md
```diff
@@ -116,75 +117,115 @@ cohortextractor run dummy generate_cohorts expectations
...
 These outputs can be [reviewed on the server](workflow-check-disclosivity.md) and released via GitHub if they are deemed non-disclosive.
 
-For those with Level 3 access:
-
-* all outputs will be saved in `E:/high_privacy/workspaces/<WORKSPACE_NAME>`
-* there's also a directory called `metadata`, containing log files for each action e.g. generate_chorts.log, run_model.log
+#### Level 3 access
 
-No data should ever be published from the Level 3 server. Access is only for permitted users, for the purpose of debugging problems in the secure environment.
+* All outputs will be saved in `E:/high_privacy/workspaces/<WORKSPACE_NAME>`
```
sebbacon, line 217: I think we should keep something _like_ "For those with Level 3 access" or similar -- anything to indicate that most people should never expect to do this and/or pay any attention to this section. Don't want to encourage people to think about using L3.

docs/workflow-make-repo.md
```diff
@@ -9,7 +9,7 @@ Click the green button that says <span style="background-color: green; color: wh
 Fill in the details:
 
 - **owner**: select your personal GitHub for testing/experimenting, or select the opensafely organisation for a bona fide OpenSAFELY-approved study. The repo can be transferred into opensafely later if needed.
-- **repository name**: This should always end in `-research` so that any back-end updates to the template can be pushed through to existing repos. The rest of the name should be something short but informative --- browse existing repo names for inspiration.
+- **repository name**: The name should be short but informative &mdash; browse existing repo names for inspiration.
```
sebbacon, line 12: Last I knew this was still a convention. I'd like to change that convention/soft requirements, but I think we need a chat with tech team first. Possibly I missed a conversation here in which case sorry!

sebbacon (approved):

docs/pipelines.md
wjchulme: I've corrected the `cohortextractor` command, pending a re-write for `opensafely`

docs/pipelines.md
```diff
@@ -116,75 +117,115 @@ cohortextractor run dummy generate_cohorts expectations
...
 ```
 
 You will also find a new folder, `metadata/`, which contains logging information about your run. If any of your actions fail, you may find clues here as to why.
 
+Each action is run in its own isolated environment in a temporary working directory. This means that all the necessary libraries and data must be imported within the script for each action &mdash; For R users, this essentially means that the R session is restarted for each action. 
+
+<details>
+  <summary>Click here for information on the exact steps that occur when each job is run locally</summary>
+  
+  1. Create a new, empty temporary directory for the job
```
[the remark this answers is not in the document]
wjchulme: I've updated using passive voice "a new temp directory is created"

docs/pipelines.md
```diff
@@ -116,75 +117,115 @@ cohortextractor run dummy generate_cohorts expectations
...
+	* select the actions you want to run by clicking the `Run` buttons
+	* if any of these actions have dependencies then they will also be run, unless their outputs already exist
 	* dependencies can be viewed by clicking the `Needs` button
 	* you can force dependencies to be run by clicking `Force run dependencies`, even if those actions have already been run
 	* when you're ready, click `Submit`.
 
 The workspace is available at `https://jobs.opensafely.org/<WORKSPACE_NAME>/`.
 You can view the progress of these actions by click the `Logs` button from the workspace, or going to `https://jobs.opensafely.org/<WORKSPACE_NAME>/logs`.
 
+<details>
```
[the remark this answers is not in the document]
wjchulme: The two `<details>` sections are slightly different so not sure it's worth using snippets here, or do you mean in another way?

docs/pipelines.md
wjchulme: Yes good point. I've re-written level 3 info as a subsection.

docs/workflow-make-repo.md
wjchulme: from https://ebmdatalab.slack.com/archives/C63UXGB8E/p1607357521090500?thread_ts=1607357374.088700&cid=C63UXGB8E

>  we have an explicit list of non-research repos, everything else is assumed research

but let me know if you still want me to revert

bloodearnest: So, we don't *rely* on the convention for security purposes, but that doesn't mean we shouldn't still prefer a convention.

wjchulme: ok have put back in

PR #67, opened 2020-12-14
Remove type column

inglesp:
This removes the empty type column from the tables of parameters.

Before:

image

After:

image


wjchulme (approved):
Very glad I didn't spend more time trying to figure out how to do this. Thank you

PR #71, opened 2020-12-15
Version bump

wjchulme:


wjchulme:
I've corrected the mix up with `docs:` versus `fix:` etc. but haven't yet properly documented how to update `requirements.txt`, because i'm stuck in `ruamel.yaml` hell and can't actually check it's working locally.

sebbacon:
Possibly you `ruamel.yaml` problems are to do with inconsistent versions?  e.g. if you bump `cohortextractor` in `requirements.txt` but `cohortextractor` itself has bumped `ruamel` then you will have a conflict in `requirements.txt`

The way to resolve this is to run `pip-compile --upgrade-package cohortextractor` 

See https://pypi.org/project/pip-tools/

PR #73, opened 2020-12-17
add description of how to use multiple by groups

annaschultze:
This PR adds a  really brief description of how to use multiple by groups in measures. Do suggest improvements, or if you think we need to add any further clarifications.


wjchulme:
This is great, thank you

PR #74, opened 2020-12-18
Document how external datasets are imported

inglesp:


wjchulme (approved):
Very interesting that it's handled differently

PR #76, opened 2020-12-18
Cohortextractor to opensafely

wjchulme:


docs/pipelines.md
```diff
@@ -107,34 +106,39 @@ When writing and running your pipeline, note that:
...
-The `cohortextractor run` command will execute one or more actions according to the `project.yaml`.
-To see its options, type `cohortextractor run --help`.
+The `opensafely run` command will execute one or more actions according to the `project.yaml`.
+To see its options, type `opensafely run --help`.
 
-For `cohortextractor run` to work:
+For `opensafely run` to work:
 
-* `cohortextractor` version `1.6.1` or higher must be installed.
+* Python version `1.7` or higher must be installed.
```
sebbacon, line 141: 3.7

sebbacon: Although perhaps we don't need to bother saying this at all; installing `opensafely` _requires_ python 3.7 and you should get an error message to this effect. If we do want to say this, it probably belongs in the "installation" section rather than here

docs/pipelines.md
```diff
@@ -143,28 +147,31 @@ For `cohortextractor run` to work:
...
+
+```bash
+opensafely run run_model
 ```
-As this action depends on the `generate_study_population` action, it will cause `generate_study_population` to be run first, followed by `run_model`.
+
 It will create the two files as specified in the `analysis/model.do` script.
 
-To force the dependencies to be run you can use for example `cohortextractor run run_model --force-run --force-dependencies`. This will ensure that both the `run_model` and `generate_study_population` actions are run, even if `input.csv` already exists.
+To force the dependencies to be run you can use for example `opensafely run run_model -f`.
```
sebbacon, line 164: I'd be tempted to use the long form here, i.e. `--force-run-dependencies`) and optionally say "`-f` for short" if you want

mkdocs.yml
```diff
@@ -44,7 +45,7 @@ nav:
         - Adding packages: requests-packages.md
         - Adding Study Definition variables: requests-variables.md
         - Updating the Documentation: requests-documentation.md
-    - Code of conduct: code-of-conduct.md
+#    - Code of conduct: code-of-conduct.md
```
sebbacon, line 48: I prefer deleting over commenting out!

docs/codelist-project.md
```diff
@@ -19,12 +17,15 @@ opensafely/current-asthma/2020-05-06
...
+Download the new codelists into the `codelist/` folder using the `opensafely` program by running
 
-You will need to add and commit these changes and push to Github.  If you don't,
-or a newer version is available than that committed, the automated tests will
-fail with an error message.
+```bash
+opensafely codelists update
+```
+
+Beware that in Windows, if one or more of these codelist files is open then this command won't be able to run.
```
sebbacon, line 26: add "; close them first"?

sebbacon (approved):
At some point I'd like us to start writing CHANGELOGs for each subproject.

I wonder if now is the time for the documentation project? Means you can point people to that file when you announce it? Just a thought - we could leave this until we've thought more systematically - maybe Dave or another dev has some opinions.

docs/pipelines.md
wjchulme: Good spot. I think it's helpful to have the requirements listed explicitly here for completeness and in anticipation of a) people who browse the docs and jump to this page first b) as a prompt for existing users whose system may have changed and who come to this page first looking for answers. 

But yes I agree -- I've made it a bit briefer and updated the python installation pages.

docs/opensafely-cli.md
```diff
@@ -0,0 +1,62 @@
...
+### `codelists`
+This will retrieve each codelist listed in `/codelists/codelists.txt` from [codelists.opensafely.org](https://codelists.opensafely.org), and save them its latest version in the same folder.
+
+Use it like this:
+```bash
+opensafely codelists update
+```
+Running it will add (or update) the codelist `.csv` files in the `codelists/` folder.
+See the [Codelist](codelist-intro.md) section for more information on codelists.
+
```
evansd, line 62: Just a note to say that I've added an `opensafely codelists check` command which will run the same check that we're going to use in the Github Actions tests. This might be useful for people, but then the solution to any issues it finds is always just to run `opensafely codelists update` so I'm not sure if it's worth mentioning at this stage.

evansd (approved):
Looks great Will, thanks for doing this. There's one small thing to change and then something to fix on my side and other than that all looks good.

docs/cohortextractor.md
```diff
@@ -40,67 +37,48 @@ If you need to install a new version, update with:
...
+      data: output/input_cohort2.csv
 ```
-cohortextractor update_codelists
-```
-Running it will add (or update) the codelist `.csv` files in the `codelists/` folder.
-See the [Codelist](codelist-intro.md) section for more information about how to create codelists.
-
-Beware that in Windows, if one or more of these codelist files is open then `update_codelists` won't be able to run.
 
+You can change the location of the outputted `.csv` file using the `--ouput-dir` option, for example `run: cohortextractor:latest generate_cohort --ouput-dir /output/cohorts`
```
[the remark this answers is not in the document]
wjchulme: I'd never understood if or how a leading `/` changed things -- thanks.

docs/opensafely-cli.md
wjchulme: I've added for completeness

PR #78, opened 2020-12-21
update protocol description

annaschultze:
This PR provides some brief placeholder text for the "protocol" section of the "Good Practices" page in the documentation. 
This previously had some internal bullet points, and I've replaced this with some text more appropriate for external collaborators. 

Long-term, I'm interested in developing this as a resource to make it easy for researchers to develop and pre-register a study protocol, and also would like to discuss whether we could recommend this a little bit more strongly (or whether we don't want to be too opinionated...).  

I wasn't 100% sure if I got all of the markdown correct (it didn't seem to render properly in RStudio), so please flag if this needs changing before merging.


wjchulme:
I think this is a great idea, but have we already agreed that we should be providing these resources for users? Are there existing materials we can point to instead? 

The markdown looks fine and there's no need to edit. The `---8<--` for the glossary and `!!!` for boxes will only be picked up by mkdocs in the actual documentation pages, not when rendering locally or in github.

annaschultze:
Good point - I've removed the mention of templates, and instead just pointed people to some of our publicly available ones in case this is for people who are new to writing protocols.

wjchulme (approved):

PR #79, opened 2020-12-28
Smooth release

wjchulme:
This separates out the redaction instructions from the release instrutions, and adds instructions for running the new [`osrelease` command](https://github.com/opensafely/output-publisher/blob/main/README.md).


sebbacon (approved):
vg!

PR #82, opened 2021-01-04
add installation instructions

wjchulme:


CarolineMorton (approved):
This looks good to me. I have not tested it by logging into the server but it looks right.

PR #103, opened 2021-01-18
Add macOS Installation Docs

ghickman:
This adds a macOS installation page taking users from no build tools to a working `opensafely` tool.

Tested on a fresh install of macOS 11.1 (Big Sur) with the os-demo-research repo.


evansd (approved):
This looks great.

Is it worth on some of the steps (I'm thinking particularly of the homebrew and docker installation steps) putting a note about roughly how long we expect it to take? I'm guessing that these two might take a while and I think it would be helpful to warn users of that.

It also might be worth adding a reassuring note at the top along the line of "yes, there are quite a few steps here but you should just be able to copy and paste the commands below and they will do everything for you".

ghickman:
> there are quite a few steps here

I did wonder about us shipping a shell script which is all of these commands just in a file?  Essentially saying "run this, put in your password when it asks you too and it should Just Work™*"

* I have just doomed this to never work.

> a note about roughly how long we expect it to take

Good point!

docs/install-macos.md
```diff
@@ -0,0 +1,86 @@
...
+
+```bash
+pipx install opensafely --python ~/.pyenv/shims/python3.9
+```
+
+Test the installation of OpenSAFELY CLI.
+This should print out the usage and available sub commands:
+
+```bash
+❯ opensafely --help                                                                                   12:01:19
```
madwort, line 42: does `❯` represent the shell prompt? Is that what a typical user on a default install would see? pre-10.15 I think it was `$` (bash) and maybe currently it's `%` (zsh) ?

docs/install-macos.md
```diff
@@ -0,0 +1,86 @@
+Open Terminal.app by clicking the magnifying glass icon in the top right of your screen.
```
madwort, line 1: Nitpicking: could we have a first line that says e.g. `# macOS Install guide`?

madwort: and maybe specify what version of macOS this guide was written for?

madwort: (it all looks like stuff that isn't going to change any time soon, but it could help with the inevitable "I tried this on my old 10.11 machine and it didn't work")

docs/install-macos.md
```diff
@@ -0,0 +1,86 @@
...
+
+Now that the Docker application is open you can click "Skip tutorial" and close the window.
+The Docker service will continue to run in the background and can be accessed from the Docker icon in your menu bar.
+
+![macOS Docker Privileges Escalation Image](./images/macos-docker-skip-intro.png)
+
+
+You're done!
+
+Now you can navigate to a research repo on your local machine and run:
```
madwort, line 82: I would break this up & put in a separate section `## running opensafely`

docs/install-macos.md
madwort: maybe e.g. a  `## requirements / intro` and a `## installing`

docs/install-macos.md
ghickman: Good point, I copied this from my local machine which is running starship, will change to `%`!

docs/install-macos.md
ghickman: What about linking to the [CLI running section](https://docs.opensafely.org/en/latest/opensafely-cli/#using-opensafely-at-the-command-line)?

docs/install-macos.md
ghickman: What would you expect to see in the intro other than explaining the guide was written with macOS 11.1?

Locally I've tried out a note:



I think it needs a bit of body on that note to set expectations but otherwise I'm not sure what else to put in there!

docs/install-macos.md
madwort: even better!

docs/install-macos.md
madwort: For me that's enough - as it gives me a reference point to a working setup - suggestions for things you could add include ", but should work with 10.14 or higher", or in future " and tested with macOS 12, 13, ...". 

(a quick scout suggests that homebrew & docker ask for 10.14, GHdesk & vscode only want 10.10)

ghickman:
@madwort – pushed fixes for all your points, could you take another run through of the document?

docs/install-macos.md
```diff
@@ -0,0 +1,87 @@
...
+
+```bash
+pipx install opensafely --python ~/.pyenv/shims/python3.9
+```
+
+Test the installation of OpenSAFELY CLI.
+This should print out the usage and available sub commands:
+
+```bash
+% opensafely --help                                                                                   12:01:19
```
madwort, line 47: typo: remove `12:01:19`

ghickman: What on… oh I copy pasted this whole block from my terminal didn't I 🤦‍♂️

madwort (approved):
🚀

Issue #108, opened 2021-01-20
Fix: Index date currently string rather than variable name in pseudo code


jogilder:
https://docs.opensafely.org/en/latest/study-def/
6th code block needs correction of last line

`age = patients.age_as_of("index_date")`

should be

`age = patients.age_as_of(index_date)`

madwort:
Thanks for the report - I've filed a PR to amend this.

wjchulme:
Closing this issue - see https://github.com/opensafely/documentation/pull/109#issuecomment-764802645

PR #109, opened 2021-01-20
Fix pseudocode typo

madwort:
* fixes https://github.com/opensafely/documentation/issues/108

### DIFF
--- a/docs/study-def.md
+++ b/docs/study-def.md
@@ -160,7 +160,7 @@ study = StudyDefinition(
             "last_day_of_year(index_date)",
         ],
     ),
-    age = patients.age_as_of("index_date"),
+    age = patients.age_as_of(index_date),
 )
 ```
 


wjchulme:
@evansd can you check this? I think both `index_date` and `"index_date"` will work, depending on where it's defined.

evansd:
Ah, yes, this is confusing. So it is possible to just define a variable as a date string e.g. `my_date = "2020-01-01"` and then use that anywhere you would normally supply a date string. But that's different from the special handling that the string `"index_date"` gets as that works with the date range thing for measures, as well as letting you do date arithmetic like `"last_day_of_month(index_date) + 10 days"`.

I'd say it's fine (and encouraged) to define variables as dates to avoid specifying the same date twenty times. But avoid using the names `index_date` or `today` as that's liable to leave to confusion.

wjchulme:
Whilst the original code block won't run on it's own (because expectations aren't defined for age), using `"index_date"` and not `index_date` is correct, because `index_date` is defined _within_ `StudyDefinition` and not as a separate global variable. So I'm going to close this.

evansd:
Oh, I wasn't actually paying enough attention. I assumed the original issue and the PR were the doing the reverse of what they actually are! (i.e. adding quotes not removing them). Yes, closing this is the right thing.

evansd:
Sorry, having looked through all of that docs page it looks like we're using a mixture of quoted and unquoted `index_date`. As it happens the unquoted version will work in the contexts in which they're used, but I think it would be less confusing if the docs used `"index_date"` everywhere.

madwort:
if you get a minute, might be worth doing a PR @evansd to reduce confusion for those who come later?

PR #120, opened 2021-02-04
Add ways to use multiple index dates

alexwalkercebm:
`common_variables` is already well documented, so I just added a clarification that you can pass different index dates. The identical study definitions idea is described in https://github.com/opensafely-core/cohort-extractor/issues/450


wjchulme (approved):

HelenCEBM:
Hi @alexwalkercebm this is very useful! But I didn't find it because was looking for it in the [index-dates](https://docs.opensafely.org/en/latest/study-def/#index-dates) section (under Defining Date Periods) - can we add a link in this section to the description you added so it's easier to find?

PR #160, opened 2021-02-22
Add documentation for new opensafely commands

bloodearnest:
- 'pull' to update docker images
- 'upgrade' to upgrade the opensafely tool itself

### DIFF
--- a/docs/opensafely-cli.md
+++ b/docs/opensafely-cli.md
@@ -21,6 +21,13 @@ To check this has installed successfully, submit `opensafely --version`.
 You should keep the tool up to date as much as possible. You can upgrade to a new version of `opensafely`, update with:
 
 ```bash
+opensafely upgrade
+```
+
+Note: if you are on an older verion of opensafely, you may need to upgrade the older way to get the upgrade command:
+
+
+```bash
 pip install --upgrade opensafely
 ```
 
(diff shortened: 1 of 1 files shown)


docs/opensafely-cli.md
```diff
@@ -20,6 +20,13 @@ To check this has installed successfully, submit `opensafely --version`.
 
 You should keep the tool up to date as much as possible. You can upgrade to a new version of `opensafely`, update with:
 
+```bash
+opensafely upgrade
+```
+
+Note: if you are on an older verion of opensafely, you may need to upgrade the older way to get the upgrade command:
```
sebbacon, line 27: Make it clearer they'll only have to do this once?

Something like:

> The above command only works with `opensafely` version `x.y.z` or newer. If you are using an older version, you will first need to upgrade it with `pip install --upgrade opensafely`

docs/opensafely-cli.md
```diff
@@ -69,3 +76,16 @@ to check if the codelist files are up-to-date with thse listed in `./codelists/c
 
 See the [Codelist](codelist-intro.md) section for more information on codelists.
 
+
+### Updating Docker Images
+
+
+To run your code on your machine, the opensafely tool uses the same docker
```
sebbacon, line 83: backtick-quote `opensafely` to make it really clear what we're talking about

docs/opensafely-cli.md
```diff
@@ -69,3 +76,16 @@ to check if the codelist files are up-to-date with thse listed in `./codelists/c
...
+images that run in the secure server environments. These are updated
+periodically, for example with new libraries installed. If you have error
```
sebbacon, line 85: > for example when new libraries are installed

sebbacon (approved):

PR #166, opened 2021-03-01
Restructure content

sebbacon:
This PR focusses on reorganising content to make content easier to find; there's very little meaningful copy change.

The biggest change is to make the relationship between actions and pipelines more explicit.

Collapsed:

image

Expanded:

image


alexwalkercebm (approved):
Looks great. I like the more nested structure.

wjchulme:
Ah great, thanks. Been hoping someone would have a better vision for this sort of thing. 

Agree more nesting is in general helpful. Content-wise though, there often isn't any meaningful delineation between "reference" material and "how-to" material (though conceptually they are different) -- there's a strong how-to element to the codelist and actions content for example. It was difficult to decide which approach was most appropriate and it's mostly ended up a mix of the two. Not sure if this means we should re-think the structure or re-think the content?

sebbacon:
Thanks both. @wjchulme I think you're absolutely right. My thought was start with the structure and the content will follow.

annaschultze:
I think re-structuring the docs is a great idea, they've expanded so much over the past few months..! Some general thoughts on the proposed structure: 

- I don't find the difference between 'reference' and 'how-to' immediately obvious; that might relate to Will's point on the content division between the two. If they're meant to cover what I think they are ('how to' guides like the user manual, reference like a reference manual for information I might come back to), I might have expected the how-to guides to come before the reference guide, and most of the content to be classed as 'how-to'. However, looking at it it does seem like most pages currently contain a bit of both, so appreciate it might not be so easy to classify these... 
- I would put Good Practice as a separate heading, as this I think is more opinion based pieces/things we're more or less strongly encouraging people to consider. 
- I would put Update the Documentation as a separate header (last), as this is more around how to contribute to the current documentation than how to get support on OpenSAFELY

PR #177, opened 2021-03-10
"Getting started" tutorial

sebbacon:
As you'll see, the motivation for this tutorial is to give us a "please do all this stuff first" check for new collaborators. It can take some people a long time to (e.g.) get permission to install docker on their work laptop, so getting it completed is a good trigger for starting the next bit of the onboarding process.

Because of images, formatting, etc, this is much easier to review when it's published. I'm tempted to publish and *then* get more feedback, but open to suggestions!


docs/getting-started.md
```diff
@@ -0,0 +1,335 @@
...
+few things. Most important is the `opensafely` tool; to install this, you must
+(currently) first install the Python programming language.
+
+### Install Python and **opensafely**
+1. [Download and run the Anaconda Python
+   installer](https://docs.anaconda.com/anaconda/install/windows/).
+2. When you've done, to verify your installation, open Anaconda Prompt by
+   clicking Start, search, or selecting Anaconda Prompt from the menu. ![Finding
+   Anaconda Prompt on Windows](./images/win-anaconda-prompt.png)
+3. Install the OpenSAFELY command line tool. Type `pip install opensafely`, and
```
CarolineMorton, line 54: Link to how to install `pip` - there are some cases on OS when you need to use `pip3` rather than `pip` which can be a gotcha

docs/getting-started.md
```diff
@@ -0,0 +1,335 @@
...
+
+### Add a chart
+
+**Every** study starts with a *study definition* like the one you just edited.
+When executed, a study definition generates a CSV of patient data.
+
+A real analysis will have several further steps after this. Each step can be
+written in [any of the programming languages supported in
+OpenSAFELY](actions-scripts.md). In this tutorial, we're going to draw a
+histogram of ages, using four lines of Python.
```
CarolineMorton, line 224: add "This can be done in R or Stata for real studies. "

mkdocs.yml
```diff
@@ -1,95 +1,91 @@
...
-        - Our philosophy: open-methods.md
-        - Security: security-levels.md
-        - Analysis workflow: workflow.md
+  - Introduction: index.md
+  - How OpenSAFELY works:
+      - Our philosophy: open-methods.md
+      - Security: security-levels.md
+      - Analysis workflow: workflow.md
+  - Getting Started guide: getting-started.md
+  - Installation guides:
```
CarolineMorton, line 9: Changing to `Installation Reference Guide` maybe - otherwise its repetitive if done tutorial

CarolineMorton (approved):
this looks good
Just a couple of comments

docs/getting-started.md
sebbacon: Oh I assumed Anaconda included `pip`. Is it as simple as `conda install pip`? 

The default Anaconda python is 3.x so doesn't that mean the default `pip` is a python 3 `pip`?

madwort: nit: I think the wording of this step could be clearer - with my new-user head on, I initially read this as "install the OS command line tool, then type pip etc", but actually you mean "install the OS command tool *by* typing pip etc into the Anaconda prompt"?!

docs/getting-started.md
```diff
@@ -0,0 +1,335 @@
...
+    [detailed installation notes](install-docker.md) may help.
+
+
+1. If you are using Windows 10 Pro, Enterprise or Education, you should [follow
+   these instructions](https://docs.docker.com/docker-for-windows/install/).
+   Otherwise, follow the instructions for [installing on Windows
+   Home](https://docs.docker.com/docker-for-windows/install-windows-home/).
+2. Starting Docker can take a while &mdash; up to 5 minutes. While it's doing
+   so, an animation runs in the notification area:<br>
+   ![image](https://user-images.githubusercontent.com/211271/72052991-14a8c000-32be-11ea-948f-575a3c84bc3b.png)
```
madwort, line 75: can we self-host this image?

docs/getting-started.md
```diff
@@ -0,0 +1,335 @@
...
+where it can be viewed by others. Your repository is automatically configured
+with tests to verify the project is runnable, each time you push.
+
+1. Open Github Desktop to view your repository. When you make changes to files in
+   your text editor and save them locally, you also see the changes in
+   GitHub Desktop. To add all changes in all files to a single "*commit*", keep the
+   checkbox at the top of the list ticked.<br>
+  ![image](images/commit-all.png)
+2. At the bottom of the list of changes, in the **Summary** field, type a short,
+   meaningful *commit message*. Optionally, you can add more information about
```
madwort, line 285: nit: what is a _commit message_? how about:
```
...In the **Summary** field, type a short meaningful description of the changes (this is called the _commit message_). You can add more information...
```

docs/getting-started.md
```diff
@@ -0,0 +1,335 @@
...
+   the change in the **Description** field. Press the blue button to make the
+   commit.<br>
+  ![image](images/commit-message.png)
+3. Click **Push origin** to push your local changes to the remote repository on
+   Github ![image](images/push-to-github.png)
+4. In a web browser, visit your repository in Github. Click on the **Actions**
+   tab ![image](images/github-actions-tab.png)
+5. You'll see a *Workflow* running with the title of your last commit. This
```
madwort, line 293: the `title`? or the `_commit message_`?

docs/getting-started.md
CarolineMorton: From memory, I think you have to install `pip` separately, although this may have changed. I think it's not included because you can install most things via `conda install` instead of `pip install`, although not what we need obviously. 

The `pip3` issue is a separate one for macOS. Because macs always come with Python 2 installed (used in operating system), `pip` defaults to python2. You either have to overwrite this for example via alias or use `pip3`.

docs/getting-started.md
```diff
@@ -51,11 +51,13 @@ few things. Most important is the `opensafely` tool; to install this, you must
 2. When you've done, to verify your installation, open Anaconda Prompt by
    clicking Start, search, or selecting Anaconda Prompt from the menu. ![Finding
    Anaconda Prompt on Windows](./images/win-anaconda-prompt.png)
-3. Install the OpenSAFELY command line tool. Type `pip install opensafely`, and
-   press ++enter++
-4. You should see some output. To check this has installed successfully, type
+3. To install the OpenSAFELY command line tool, you first need to install a tool
+   called `pip`.  Do this by typing `conda install pip` and pressing ++enter++.
+4. Now you can install the comment line tool, by typing `pip install opensafely`, and
```
madwort, line 56: typo `comment line`

PR #182, opened 2021-03-12
Bmk tutorial suggestions

brianmackenna:
This tutorial is great and exactly what is needed. I have run to completion (with exception that I used windows powershell in place of anaconda).

 I have more suggestions about additions (e.g. using codelists) but sending this now with a tweak on `project.yaml` file so others can start using the tutorial.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -256,7 +256,7 @@ actions:
         cohort: output/input.csv
 
   describe:
-    run: python:latest python analysis/describe.py
+    run: python:latest python analysis/report.py
     needs: [generate_study_population]
     outputs:
       moderately_sensitive:
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -154,7 +154,8 @@ generate_study_population: Cleaning up container and volume
 The final line tells you a file of (random) patient data has been created at
 `output/input.csv`, and that it would normally be considered highly sensitive
 data. What you see here reproduces exactly what would happen on a real, secure
-server, but with *dummy data* which contains no real patient information.
+server, but with *dummy data* which contains no real patient information. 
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -154,7 +154,8 @@ generate_study_population: Cleaning up container and volume
 The final line tells you a file of (random) patient data has been created at
 `output/input.csv`, and that it would normally be considered highly sensitive
 data. What you see here reproduces exactly what would happen on a real, secure
-server, but with *dummy data* which contains no real patient information.
+server, but with *dummy data* which contains no real patient information. 
+Remember this is a blank study, read on below to generate *dummy data* values.
```
sebbacon, line 158: I'm not convinced about this addition - there are quite a few details I've missed out here because I want to keep the word count as short as possible. My hypothesis is that someone who looks at the file and wonders "why is there only one column", will not be confused enough to stop; they'll keep going, and then the next steps make this obvious?

madwort, line 158: Good idea, although re-reading it in context I'm not sure we've really signposted that this is a blank study - so instead of "Remember...", might be better as "Because we haven't modified the template blank study yet, this file is empty - we'll generate some *dummy data* values in the next section" - wdyt?

madwort: heh, crossed in the post @sebbacon

sebbacon: i.e. I only think you were confused because you already know what you were looking for

brianmackenna: I think that is a much better wording suggestion @madwort  thanks!

 @sebbacon I know I'm not target audience but I was confused enough to stop and wonder where the dummy data

sebbacon (approved):

PR #200, opened 2021-03-25
Add feedback from Amir

sebbacon:


docs/getting-started.md
```diff
@@ -72,15 +72,22 @@ few things. Most important is the `opensafely` tool; to install this, you must
...
-   click system tray docker icon; select "settings"; select "shared drives".
+3. If you have ended up with the *Hyper-V* backend, then when Docker has
+   finished starting up, you will need share your hard drive with Docker: click
+   system tray docker icon; select "settings"; select "shared drives".  This
+   setting does not exist in the *WSL 2* backend.
 4. Test Docker and opensafely work together. Open an Anaconda Prompt, and run
-   `opensafely pull`.  The first time you run it, this may take a long time,
-   depending on your network connection. It is downloading a reproducible
-   environment identical to that installed in the OpenSAFELY secure
+   `opensafely pull --force`.  The first time you run it, this can take a very
```
bloodearnest, line 88: I would remove this step entirely. `opensafely run` will download any images it needs.  There's no need to force the user to download every image (~10Gb) when they are only going to use `cohortextractor` image (~1GB) in this getting started guide.

If you really want to pre-load it, do `opensafely pull cohortextractor`

bloodearnest (changes requested):

ghickman:
I've updated this to take into account feedback, some tweaks to make editing (particularly inserting into lists) easier, and based it on the current state of `master`.  One of the paragraphs from the original changeset has already been rolled in (noting there are two backends).

@StevenMaude – you're probably best to review this?

StevenMaude (approved):

PR #202, opened 2021-03-26
Restructure study def into subsections

sebbacon:
Also adds some new brief intro text to each page, and some new copy re. thinking about null dates


wjchulme:
Ha, I have a branch that does this for dummy data. With some other (incomplete) simplifications to the whole section given what we now know about how study definitions are typically used. I'll rebase once this is merged.

wjchulme (approved):

PR #207, opened 2021-03-30
Study def tidy

wjchulme:
This PR simplifies the study definition pages, in particular making the explanation of `population` more consistent with how it's typically used.


evansd (approved):
Yep, looks good!

PR #213, opened 2021-04-08
Updates to working with dates

HelenCEBM:


sebbacon (approved):

docs/study-def-dates.md
```diff
@@ -71,6 +78,8 @@ actions:
         cohort: output/input-2020-01-01.csv
 ```
 
+This can also be used to define a range of dates over which to run the study_definition, usually when working with [Measures](https://docs.opensafely.org/en/latest/measures/).
```
madwort, line 81: I think the target of this link should be a local relative one `(measures.md)`, rather than the absolute url `(https://docs.opensafely.org/en/latest/measures/)`

madwort (changes requested):
url needs fixing (it could silently break deployment in the future), otherwise looks good

docs/study-def-dates.md
HelenCEBM: Thanks, good suggestion - I've now done this, hopefully correctly!

evansd (approved):
This looks good @HelenCEBM and I think actually covers the change I made recently. Or at least, the change I made recently now makes this correct.

We could explain that dates in EMIS default to NULL but that we translate them to 1900-01-01 behind the scenes. But that feels like extraneous detail at this point.

PR #217, opened 2021-04-08
Explain force run

iaindillingham:
Two commits, which hopefully explain the force run argument. I'm going to assume that as it was helpful to me to write this explanation, then it will be helpful to other people to read it 🙂 

A couple of things tripped me up, when walking this path. The first was that the Getting Started Guide used `--force`. I *think* argparse expands this silently, but I took it that the action itself was forced as well as its dependencies. The second was that I didn't realize that specifying an action always runs that action. Again, because of the first trip up, I took it that specifying an action without the argument was a no-op.

Thanks @bloodearnest for talking this through, today.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -216,7 +216,7 @@ between these two dates*"; lines 14-15 "*Give me a column of data corresponding
 to the age of each patient on the given date*"; and lines 16-18 "*I expect
 every patient to have a value, and the distribution of ages to match that of the
 real UK population*"
-1. In an Anaconda Prompt, run `opensafely run run_all --force`. A new file will
+1. In an Anaconda Prompt, run `opensafely run run_all --force-run-dependencies`. A new file will
    be created in the folder `output/input.csv`. Open that file (using Visual
    Studio Code, or Excel) and you'll see it now contains an age for 1000
    randomly generated patients.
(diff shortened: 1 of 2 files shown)


iaindillingham:
We might wish to consider whether the documented behaviour is desired. As an example, I have a study definition with three actions: `generate_study_population`, `rotate_study_population`, and `generate_summary_notebook`.

First, I execute `opensafely run run_all` which generates outputs for each action.

* If I then execute the same command, then the CLI reports `All actions already completed successfully` and the outputs for each action are not regenerated.
* Alternatively, if I then execute `opensafely run generate_study_population rotate_study_population generate_summary_notebook`, which to my mind is conceptually the same as running all the actions, then the outputs for each action are regenerated.

@bloodearnest thought that if a user specified one or more actions explicitly, then they would expect the outputs for each action to be regenerated. This is a valid assumption! Perhaps it's worth keeping an ear open for any user feedback, to check it.

sebbacon:
Have added @evansd as reviewer, I'm not sure I'll get time today, sorry!

evansd (approved):
Thanks for this!

I had no idea that argparse accepted abbreviated versions of argument names like that. It looks like we can disable that behaviour now, and I think we should:

PR #224, opened 2021-04-21
Correct a link for the macOS install guide

StevenMaude:
From the Getting Started page. This might be a frustration for Mac
users, particularly as the much less obvious alternative route for the
Mac instructions is to click multiple times through the sidebar.

This was flagged by `mkdocs build --strict`.


madwort (approved):
Thanks for the catch!

StevenMaude:
Current site points to: `https://docs.opensafely.org/en/latest/getting-started/docs/install-macos.md` which gives a 404 error.

After this PR is merged, there are no more warnings from `mkdocs build --strict` :partying_face:

PR #228, opened 2021-04-21
Add a simple build check workflow

### LINKED ISSUE #223: Add a build check on pull request
This would flag issues to a PR author and reviewer(s) before attempting to deploy a new version of the documentation

StevenMaude:
Fixes #223.

Add a workflow to check the documentation build on pull request.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,6 +14,8 @@ jobs:
 
       - name: Set up Python
         uses: actions/setup-python@v2
+        with:
+          python-version: '3.7'
 
       - name: Install dependencies
         run: pip install -r requirements.txt
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,22 @@
+---
+name: 'build-check'
+
+# yamllint disable-line rule:truthy
+on:
+  pull_request:
+
+jobs:
+  build-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+
+      - name: Install dependencies
+        run: pip install -r requirements.txt
+
+      - name: Check site builds
+        run: mkdocs build --strict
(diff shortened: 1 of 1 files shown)


StevenMaude:
This works as proof-of-concept, but I have some thoughts and questions for refining:

* Does this want to apply to PRs to all branches, or just main/master?
* I specified the Python version as 3.7 as [referenced by @madwort](https://github.com/opensafely/documentation/issues/223#issuecomment-823944057) but Python 3.8 is used in the [deploy workflow for GitHub Actions](https://github.com/opensafely/documentation/blob/ebe012148f7b5a0edeb1fb152915fcbce27971fb/.github/workflows/publish.yaml).
  * Should this or the `publish` workflow change their Python version?
* I've used `mkdocs build --strict` as this previously flagged useful fixes.
  * Should this or the `publish` workflow change to use or not use `--strict`?
* Not sure whether a build check on pull request should bother using GitHub's caching.
  * In theory, I guess it's possible someone unintentionally or maliciously adds a ton of useless requirements to the cache, particularly as this is a public repository and anyone can make a pull request.
* An alternative might be a shared [composite run steps](https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action) action — used by both `build` and `publish` — that lives in this repository.
  * Not sure this is worth it to avoid duplication as it's a very simple script.
  * But if we used caching for the build check, we could remove the duplication. I think a composite run action could contain everything in the `deploy` workflow, up to deploy. Then the `deploy` workflow just has a final additional step.

madwort:
@StevenMaude I think it's worth pointing out that the deploy workflow is currently "not used" - it deploys to https://opensafely.github.io/documentation/ which is a completely separate deployment to https://docs.opensafely.org/en/latest/ - the latter operating by a hook on readthedocs.org which gets notified on merge to master & builds the site using readthedocs own workflow - probably we should remove the deploy workflow from this repository

madwort:
or, alternatively, we could migrate from readthedocs.org to github-pages, in which case your ideas about the deploy workflow sound great!

madwort (approved):

Issue #42, opened 2020-11-27
Documentation for Codelist Functions


annaschultze:
A couple of functions - `combine_codelists` and `filter_codes_by_category` are listed in the documentation for the study_definition; I wasn't sure how to find out what these did. 

I gather from @alexwalkercebm that these allow you to combine multiple codelist to create a single variable, or use codelists with a "category" (I assume an additional column?), which is really useful.

I wasn't sure if these were already documented somewhere and I just could not find it, if not it may be worth adding to the documentation (including the codelist section, as it might have implications for how people develop and organise their codelists). Happy to propose edits and open a PR if relevant/useful.

alexwalkercebm:
Another variant on the use of `filter_codes_by_category` in this Slack chat:

> john tazare Today at 16:41
> @inglesp @alex Apologies if I've missed where this was previously explained.
> I want to add a column that categorises an existing codelist (e.g. to distinguish incident/history of codes), aside from uploading a separate codelist is there a better way to do this ?
> MI is a good example:
> https://codelists.opensafely.org/codelist/opensafely/myocardial-infarction/2020-06-25/
> 8 replies
> 
> inglesp  5 minutes ago
> For now (and the foreseeable future) you'll need to upload a new version of the codelist
> :+1:
> 1
> 
> 
> john tazare  4 minutes ago
> Thanks - wanted to check before potentially duplicating lists
> :+1:
> 1
> 
> 
> alex  4 minutes ago
> you could do it in a similar way to that in the type categorisation of this one. https://codelists.opensafely.org/codelist/opensafely/incident-venous-thromboembolic-disease/2020-09-30/#full-list
> :point_up:
> 1
> :+1:
> 1
> 
> 
> alex  2 minutes ago
> it's then imported as a category column https://github.com/opensafely/post-covid-thrombosis-research/blob/master/analysis/codelists.py#L39
> 
> alex  1 minute ago
> and filtered like this https://github.com/opensafely/post-covid-thrombosis-research/blob/master/analysis/common_variables.py#L15

sebbacon:
For reference - all the above should get added to docstrings [here](https://github.com/opensafely/cohort-extractor/blob/f20d096f5f64862208767ce1c13f06b15307bfde/cohortextractor/codelistlib.py#L44) and then included in our API documentation in mkdocs

PR #229, opened 2021-04-22
codelist_clarification

annaschultze:
This PR adds text to the `codelist-creation` section of the documentation describing: 

- that users can only upload csv files with two columns under their own profiles 
- situations when they might want to use more than two columns 
- advice to contact the opensafely core team in order to enable codelists with more than two columns to be uploaded  

The description of ```filter_codes_by_category``` is an outstanding issue (#42), I don't think this is currently describe in the docs.


inglesp (approved):
VG.

PR #231, opened 2021-04-22
Remove workflow publishing to GitHub Pages

StevenMaude:
We're not using this right now as we use a readthedocs-based approach
that deploys the main branch. See [@madwort's
comment](https://github.com/opensafely/documentation/pull/228#issuecomment-823986042).
The presence of this is therefore potentially confusing.

In Tech Team meeting 2021-04-22, it was generally agreed that GitHub
Pages might not be the best place, so it follows that we should remove
this workflow. If that view changes, this commit could simply be
reverted.


madwort (changes requested):
looks good, although we could probably remove some bits from `README.md` at the same time

StevenMaude:
> although we could probably remove some bits from README.md at the same time
Good catch.

StevenMaude:
We should also :wastebasket: `gh-pages` branch and ensure GitHub Pages is disabled (which might happen if the branch is deleted, maybe?).

madwort:
>We should also 🗑️ gh-pages branch and ensure GitHub Pages is disabled (which might happen if the branch is deleted, maybe?). 
yes definitely, I'll do that now

madwort (approved):

PR #243, opened 2021-05-05
to add the output-checking template

amirmehrkar:
ive copied the word file into the opensafely google folder and make a public link - hopefully that will work


LFISHER7 (approved):
I have added a word doc of the review form instead of linking to the google doc

PR #253, opened 2021-06-08
Apcs documentation expansion

HelenCEBM:


docs/dataset-apc.md
```diff
@@ -20,6 +21,214 @@ Essentially, a spell is a collection of episodes (often just one episode) denoti
...
+
+
+### `patient_classification`
+
+#### Most useful for:
+- excluding regular admissions (i.e. where patient is admitted e.g. every week for chemotherapy or dialysis) 
+- excluding mother and baby admissions (where using delivery facilities only)
+- distinguishing elective day case admissions from ordinary admissions (day case admissions are for planned procedures not requiring overnight stays)
+
+An advantage of this field is that it reflects what actually happened during the spell as well as what was planned to occur - if treatment was not limited to the planned course (e.g. patient was expected to have their routine chemotherapy but e.g. illness arose which required further treatment and an overnight stay), this value defaults to `1`.
```
wjchulme, line 156: > This value defaults to `1`. 

Does this mean that if `patient_classification` is not specified, then only ordinary admissions are picked up? Has this always been the case?

wjchulme (approved):

docs/dataset-apc.md
HelenCEBM: No this refers to the coding, not the cohortextractor - will clarify

StevenMaude:
I looked at the Cloudflare Pages failure here because I wondered why that failed, but the GitHub Actions check (should be the same) passed.

When the site was being built, it couldn't find the `images/create-repository-owner.png` for `getting_started.md` which is unrelated to the changes here. Probably due to the GitHub/Fastly outages earlier.

HelenCEBM:
@StevenMaude thanks - it's still failing, do you think that's a problem?

StevenMaude:
So this is my mistake at making a quick assumption :. I got confused because of the Fastly outage last week and assumed there was some weirdness there :flushed:

Short answer: the fix is to rebase this branch on the main branch, so that it's up-to-date.

Slightly longer answer: the reason for the failure on GitHub Actions, but passing on Cloudflare Pages is a little convoluted. I'll try and explain, because it took me a few minutes to figure out:

1. @HelenCEBM started working on this branch back at the start of the March.
2. This was before we checked with `mkdocs build --strict` which fails on any warnings. A couple of issues also got fixed as a result of using `--strict` including a broken image link in the Markdown that causes the failure here. These fixes are now in the `master` branch.
3. Cloudflare Pages checks out the most recent commit of *this* branch. The most recent commit of this branch still contains the broken link. If you checkout this pull request branch, it still fails `mkdocs build --strict`. 
4. The default [GitHub Actions checkout action behaviour](https://github.com/actions/checkout/issues/15) is that *it creates a merge commit of the pull request branch and the branch being merged*. (I realised this on looking at the Actions run logs and not being able to find the commit being tested in the repository.) For the merge commit, the docs **don't** have this broken link because it is fixed in the main branch.

PR #264, opened 2021-06-11
Fix Python syntax in examples

inglesp:
I've formatted each block of Python code with black, and then added a
few line breaks for clarity.


evansd (approved):
Black really has made life easier, hasn't it?

PR #266, opened 2021-06-14
Updates for new users

rebkwok:
Some suggested changes from following the getting started guide (gitpod and local) and the first few docs referenced from there

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -168,6 +168,8 @@ account, for developing your own study:
    some setup in background. Wait about 1 minute, then reload the page, and you
    should see the README displayed now reflects the name you gave to the new
    repository.
+   
+   If you see `${GITHUB_REPOSITORY_NAME}` in your README, the repo is not yet initialised, wait a few seconds longer and reload. 
 
 
 ## 3. Setup the required software
@@ -266,14 +268,15 @@ account, for developing your own study:
     !!! note
         On Windows, installing Docker is usually
         straightforward, but can sometimes be complicated, depending on your exact
-        version and configure of Windows. If you run into problems, our more
+        version and configuration of Windows. If you run into problems, our more
         [detailed installation notes](install-docker.md) may help.
 
 
-    1. If you are using Windows 10 Pro, Enterprise or Education, you should [follow
-       these instructions](https://docs.docker.com/docker-for-windows/install/).
-       Otherwise, follow the instructions for [installing on Windows
-       Home](https://docs.docker.com/docker-for-windows/install-windows-home/).
+    1. Follow the instructions [here](https://docs.docker.com/docker-for-windows/install/) 
+       to install Docker.  If you are using Windows 10 Pro, Enterprise or Education, you should 
+       follow the instructions for Hyper-V backend and Windows containers.  If you are using
+       Windows Home or an earlier build of Pro or Education, follow the instructions for 
+       WSL 2 backend.
        Unfortunately, we've had reports that installing in Windows Home can
        be very challenging. Please let us know if you can help us [improve
        the documentation](requests-documentation.md) here.
@@ -281,7 +284,7 @@ account, for developing your own study:
        so, an animation runs in the notification area:<br>
        ![image](images/docker-windows-starting.png)
     3. When Docker has finished starting up, share your hard drive with Docker:
-       click system tray docker icon; select "settings"; select "shared drives".
+       click the system tray docker icon; select "settings"; select "shared drives".
     4. Test Docker and opensafely work together. Open an Anaconda Prompt, and run
        `opensafely pull cohortextractor`. This will pull down the OpenSAFELY
        cohortextractor images, which can be used to run actions in your study.  The
@@ -291,7 +294,7 @@ account, for developing your own study:
 
 ## 4. Set up your first study
 
-In a previous step, you previously created a copy of the research code
+In a previous step, you created a copy of the research code
 in GitHub. You need to have a copy, or "clone", of that code to develop your own
 study.
 
@@ -336,7 +339,8 @@ study.
 
 ### Run your first study
 
-Now you're ready to run your first study. Run:
+Now you're ready to run your first study. Ensure your current directory is your newly-cloned 
+study repository, and run:
 
 ```sh
 $ opensafely run run_all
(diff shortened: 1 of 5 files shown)


docs/getting-started.md
```diff
@@ -266,22 +268,23 @@ account, for developing your own study:
...
-        version and configure of Windows. If you run into problems, our more
+        version and configuration of Windows. If you run into problems, our more
         [detailed installation notes](install-docker.md) may help.
 
 
-    1. If you are using Windows 10 Pro, Enterprise or Education, you should [follow
-       these instructions](https://docs.docker.com/docker-for-windows/install/).
-       Otherwise, follow the instructions for [installing on Windows
-       Home](https://docs.docker.com/docker-for-windows/install-windows-home/).
+    1. Follow the instructions [here](https://docs.docker.com/docker-for-windows/install/)
```
rebkwok, line 275: The old link for installing on windows home just redirects to the general windows install page now

docs/getting-started.md
```diff
@@ -489,58 +493,111 @@ real UK population*"
...
 **Every** study starts with a *study definition* like the one you just edited.
 When executed, a study definition generates a CSV of patient data.
 
-A real analysis will have several further steps after this. Each step can be
-written in [any of the programming languages supported in
+A real analysis will have several further steps after this. Each step is defined 
+in a separate file, and can be written in [any of the programming languages supported in
 OpenSAFELY](actions-scripts.md). In this tutorial, we're going to draw a
-histogram of ages, using four lines of Python.
+histogram of ages, using either four lines of Python or just a few more lines of R.
```
rebkwok, line 507: It wasn't immediately obvious to me how you'd combine different languages, so the same example in python and R felt like it'd be useful

docs/index.md
```diff
@@ -18,7 +18,9 @@ OpenSAFELY is a collaboration between the DataLab at the University of Oxford, t
 
 **Only approved users can execute code against the real OpenSAFELY database**.
 
-However, all the OpenSAFELY-specific tools needed to define analysis datasets, generate dummy data, and run analysis scripts in a computational environment that mimics the secure environment where real analyses are run, are available now for use by anyone.
+However, anyone can access and use all the OpenSAFELY-specific tools needed to define analysis datasets, generate dummy data, and run analysis scripts in a computational environment that mimics the secure environment where real analyses are run.
+
+All steps in the [Getting Started](getting-started.md) guide can be run by anyone interested in learning more about how code is written and executed in OpenSAFELY.
```
rebkwok, line 23: I think it's helpful to be a bit more explicit about the fact that you can try things out without being an approved user

CarolineMorton: agreed

docs/study-def-expectations.md
```diff
@@ -70,7 +80,21 @@ The following options are currently available for dummy data:
...
+* used for the distribution of date values, with either:
+  * `"exponential_increase"`
+  * `"uniform"`
+    
+* or for non-date values:
+  * `"universal"`: indicates every patient is expected to have a value (i.e. an alias for `incidence=1`)
+    
+`"distribution"`(numeric variables) currently has two possible options:
+* `normal`
+* `population_ages`: samples from the distribution of ages in the UK taken from [the Office for National Statistics](https://www.ons.gov.uk/peoplepopulationandcommunity/populationandmigration/populationprojections/datasets/tablea21principalprojectionukpopulationinagegroups).
```
CarolineMorton, line 100: this is great!

CarolineMorton (approved):
Much clearer.

docs/getting-started.md
wjchulme: We've a preference to use more decriptive links (ie, avoiding "here"). So this would be better as eg "follow [the Docker for Windows installation instructions](link...)."

docs/getting-started.md
```diff
@@ -390,8 +394,8 @@ information in the next section.**
 
 === "Windows (local)"
 
-    The newly created output files are created in the local copy of the
-    repository in the `output/` directory.
+    The newly output files are created in the local copy of the
```
wjchulme, line 397: newly -> new

docs/getting-started.md
wjchulme: I could make it 4 lines of R if brevity wins in this context?

wjchulme: But I like having the two options!

docs/getting-started.md
```diff
@@ -569,8 +626,9 @@ the new commit.
 
     <h3>Allow Gitpod to be able to push your changes to GitHub</h3>
     
-    1. When logged into Gitpod, visit their [Settings
-       page](https://gitpod.io/settings).
+    1. When logged into Gitpod, use [this link](https://gitpod.io/settings)
```
wjchulme, line 629: again, this doesn't quite match the hyperlink style we've tried to stick to elsewhere

docs/study-def-expectations.md
```diff
@@ -5,7 +5,7 @@ OpenSAFELY requires you to define *expectations* in your study definition: these
 ## Defining `return_expectations`
 
 Every variable in a study definition must have a `return_expectations` argument defined (with the exception of the `population` variable).
-This defines the general shape or distribution of the variables in the dummy data used for developing the code.  It is currently relatively unsophisticated; each variable is generated independently of all others. This is sufficient for testing that it is possible to run your study from start to finish in most cases, but sometimes not. You can find (and contribute to!) discussions on improving the [dummy data framework](https://github.com/opensafely/cohort-extractor/issues/221).
+This defines the general shape or distribution of the variables in the dummy data used for developing the code.  It is currently relatively unsophisticated; each variable is generated independently of all others. This is sufficient for testing in that it is possible to run your study from start to finish in most cases, but sometimes not. You can find (and contribute to!) discussions on improving the [dummy data framework](https://github.com/opensafely/cohort-extractor/issues/221).
```
wjchulme, line 8: Not sure why "in" has been added. The meaning is "In most cases, dummy data is good enough to test that it is possible to run your study from start to finish" perhaps this phrasing is better all round?

docs/study-def-expectations.md
wjchulme: thank you, much clearer

wjchulme (changes requested):

docs/getting-started.md
rebkwok: I mean, this is really the limit of my R abilities.  If you would like to tell me the 4 lines of R, I'd be happy to update it.  I could make it 4 lines of R, but only by removing line breaks :)

docs/study-def-expectations.md
rebkwok: Ah yes - the sentence just didn't read properly in my head.  I think the alternative phrasing is clearer.

wjchulme (approved):

PR #268, opened 2021-06-15
Fix ordered list numbering

iaindillingham:
Two commits; the first is the substantive change and the second is the cosmetic change. See the commit messages for the rationale.


inglesp (approved):

StevenMaude:
Semantic line break idea is a new one on me :thinking::slightly_smiling_face:

iaindillingham:
@inglesp pointed me to the website; I think the concept is excellent. It works well for both command-line and IDE peeps, and ensures commits are easier to reason about IMHO.

PR #278, opened 2021-06-16
Add initial backend access docs for emis

bloodearnest:


rebkwok (approved):
This has been around for a while, but it looks good to me. Is there a reason not to merge it?

bloodearnest:
Closing this as we don't want to add EMIS docs to public docs yet, as it's not accessible.

PR #295, opened 2021-06-23
Fix macOS Install Instructions

### LINKED ISSUE #280: Improve Python installation guidance on macOS
@CarolineMorton and I had a feedback session with a user today. They couldn't complete the Getting Started guide because they encountered issues with Python. The `pyenv` steps seems incomplete and we're not sure it's entirely necessary by default.

ghickman:
`pipx` installs libraries into their own venvs, exposing their binaries in `~/.local/bin`, but we still need to put that on the $PATH.  We've opted to take this route instead of `pipx run <tool>` to avoid tying ourselves to pipx across multiple platforms.

Fixes #280

### DIFF
--- a/docs/install-macos.md
+++ b/docs/install-macos.md
@@ -20,6 +20,18 @@ Once homebrew is installed use it to install [pipx](https://pipxproject.github.i
 brew install pipx pyenv
 ```
 
+Add pipx to your path with:
+
+```bash
+echo 'export PATH="$HOME/.local/bin/:$PATH"' >> ~/.bashrc
+```
+
+and reload your shell environment with:
+
+```bash
+source ~/.bashrc
+```
+
 Next, install [Docker for Mac](https://docs.docker.com/docker-for-mac/install/), [GitHub Desktop](https://desktop.github.com/), and [Visual Studio Code](https://code.visualstudio.com/):
 
 ```bash


StevenMaude (approved):
Is just adding to the user's default path all that is required? :facepalm: after spending 10 minutes last week frantically searching through the documentation trying to figure this out :grin:

It looks good, though I can't verify it. It would be nice if another macOS user can confirm, but I don't think that's essential.

ghickman:
@StevenMaude – unfortunately yes 😩  pipx is effectively installing into unactivated virtualenvs and symlinking the binaries they install to a central location you can put on the path.  It looks like their docs have shifted to prefer using `pipx run <tool>` now which is frustrating!

ghickman:
@CarolineMorton – maybe you could test this out since you saw the user struggling too?

docs/install-macos.md
```diff
@@ -20,6 +20,12 @@ Once homebrew is installed use it to install [pipx](https://pipxproject.github.i
 brew install pipx pyenv
 ```
 
+Add pipx to your path with:
+
+```bash
+echo 'export PATH="$HOME/.local/bin/:$PATH"' >> .bashrc
```
madwort, line 26: would you need to restart your shell session for this to take effect?

ghickman: ah, of course, we'll need to source bashrc for it take effect, thanks!

ghickman: Fixed and rebased, also caught that I wasn't explicitly setting the path to `.bashrc`.

madwort: much better, thanks!

PR #297, opened 2021-06-28
Simple suggestions for repo publication

sebbacon:


mkdocs.yml
```diff
@@ -56,6 +56,7 @@ nav:
       - Developing a protocol: protocol.md
       - Using Git effectively: git-workflow.md
       - Code reviews: code-reviews.md
+      - Publishing your Gihub Repository: publishing-repo.md
```
iaindillingham, line 59: We're missing a "t". We also write "GitHub" rather than "Github" elsewhere.

iaindillingham (approved):
I've fixed a typo, just because it's so easy to do in the UI. We might wish to replace references of Github to GitHub, for consistency with other pages, but that's not a blocker.

PR #305, opened 2021-07-08
clarify that --output_dir for generate_measures also defines the input directory

Jongmassey:
If either the generate_cohort or generate_measures steps have an --output_dir defined, then they both must have this configured identally for generate_measures to pick up the files. I've tried my best to clarify this in a couple of sentences.

### DIFF
--- a/docs/measures.md
+++ b/docs/measures.md
@@ -170,7 +170,7 @@ This will prevent it from recalculating the measure for any months or weeks whic
 However the final step, which combines output across time periods, will always be run.
 
 This command accepts an `--output-dir` argument, which defines the directory in which the output measure files will be created. 
-If this is configured, the command will also look for the input files in this directory, and so both this and the generate_cohort command in step 2 must be configured identically.
+If the `--output-dir` argument is configured, `generate_measures` expects the input cohort files to also be in the output directory. To ensure this, include the same `--output-dir` argument in the previous `generate_cohort` step.
 
 ## Putting it all together in a pipeline
 
--- a/docs/measures.md
+++ b/docs/measures.md
@@ -169,6 +169,9 @@ This command also respects the `--skip-existing` flag.
 This will prevent it from recalculating the measure for any months or weeks which have already been calculated.
 However the final step, which combines output across time periods, will always be run.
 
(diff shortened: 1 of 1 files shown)


docs/measures.md
```diff
@@ -169,6 +169,9 @@ This command also respects the `--skip-existing` flag.
 This will prevent it from recalculating the measure for any months or weeks which have already been calculated.
 However the final step, which combines output across time periods, will always be run.
 
+This command accepts an `--output-dir` argument, which defines the directory in which the output measure files will be created. 
+If this is configured, the command will also look for the input files in this directory, and so both this and the generate_cohort command in step 2 must be configured identically.
```
StevenMaude, line 173: This is OK as is :slightly_smiling_face: The below is just for polish more than anything.

nit 1: The "generate_cohort" here would be better inside backticks for consistency :sparkles: (That formatting looks to be the convention throughout the measures documentation.)

nit 2: So I don't have much context here. But I did manage to pick up the meaning, so this is OK as is. However, the second sentence might be slightly easier to follow by rewording to something like:

> If you configure the `--output-dir` argument, `generate_measures` expects the *input* cohort files to be in that same directory. To ensure this, include the same `--output-dir` argument in the previous `generate_cohort` step.

It is tricky to phrase the message here cleanly, though :thinking:

(This might also be better suggested under the `generate_cohort` step. That's assuming someone reads top to bottom. Instead of the reader doing a step, and then seeing they should have done that differently. But I don't have too strong opinion an opinion on that.)

StevenMaude: I'll approve and leave up to @Jongmassey as to if he wants to edit further. I'd suggest the `generate_cohort` fix is a quick one.

StevenMaude (approved):

docs/measures.md
Jongmassey: Yeah, I was struggling with the phrasing. Everything else seems to be in passive-ish voice so have phrased accordingly

Jongmassey:
@StevenMaude can I go ahead and merge or does it need another pair of eyes? I picked @madwort purely because he appeared to be another common reviewer of docs PRs

PR #321, opened 2021-07-15
Pin python to 3.7

madwort:


StevenMaude (commented):
Are there reasons to not specify the latest 3.8 (currently 3.8.11)?

Keeping this up-to-date is another chore, admittedly, but necessary because of dependabot. This version might also influence what version a developer runs via pyenv; as I learned from @ghickman today.

Related point: The GitHub Actions workflow uses Python 3.7. I have no idea what Cloudflare Pages uses in deploy. It would be good if all these were consistent.

(I think all this is possibly a general maintenance issue with some of our repositories. Not necessarily too important, but the kind of thing that can catch us out in future.)

madwort:
good question - picked this because I think it's a known-good version - checked the CloudFlare logs for this PR & it says 
```
14:12:55.156 | Python version set to 3.7
```
so will revise to 3.7.last

StevenMaude:
There might not be good answers right to this now. "It makes Dependabot work here" might be a good enough response :slightly_smiling_face: — especially as merging a Dependabot PR should require the docs build check to pass.

It does highlight some general problems of managing all this Python version stuff.

StevenMaude:
> so will revise to 3.7.last

Sounds good :+1:

I don't know why this repository is on 3.7 and not 3.9 actually. Definitely outside scope of this PR though :slightly_smiling_face:

madwort:
it's an ENVVAR set in the config panel on CloudFlare pages
```
PYTHON_VERSION | 3.7
```

StevenMaude (approved):
Apparently, there's [3.7.*11*](https://www.python.org/downloads/release/python-3711/), if you want to bump to that, otherwise it's fine for now, given the use of it (which is primarily for Dependabot). I'll approve regardless and leave that up to you.

PR #327, opened 2021-07-15
Use mkdocs 1.2.1

madwort:
I suspect that upgrading mkdocs & mkdocstrings at the same time might fix CI - let's see.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,4 +26,4 @@ jobs:
           python3 -m pip install -r ./requirements.txt
 
       - name: Check site builds
-        run: mkdocs build --strict
+        run: mkdocs build --strict -f mkdocs.test.yml
--- a/mkdocs.test.yml
+++ b/mkdocs.test.yml
@@ -1,0 +1,2 @@
+INHERIT: ./mkdocs.yml
+site_url: https://docs.opensafely.org/
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,6 @@ chardet==3.0.4
 click==7.1.2
     # via
     #   mkdocs
-    #   nltk
     #   pip-tools
 cryptography==3.4.4
     # via pyopenssl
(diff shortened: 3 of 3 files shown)


StevenMaude:
It needs:

```
site_url: https://docs.opensafely.org
```

in mkdocs.yml

Then this will cause:

```
  File "templates/python/material/parameters.html", line 14, in top-level template code
    <td>{{ parameter.description|convert_markdown }}</td>
TypeError: do_convert_markdown() missing 1 required positional argument: 'heading_level'
```

madwort:
thanks, added a separate mkdocs.test.yml, because I wanted to avoid breaking links on the cloudflare test deploys - it gives the error that the mkdocstring pr gives on its own, though - maybe we can fix that there & do mkdocs-1.2 afterwards

StevenMaude:
This is all resolved now via #332 and #333. In mkdocs 1.2.1 there was a requirement to specify `site_url`: this requirement was [removed in 1.2.2](https://github.com/mkdocs/mkdocs/pull/2490).

***

NB: `site_url` is to specify a canonical URL. We possibly wouldn't then want `https://docs.opensafely.org` as `site_url` for local builds or Cloudflare Pages previews, but just for the deployed site. 

Cloudflare Pages lets you discern [production from non-production branch](https://developers.cloudflare.com/pages/platform/build-configuration). If we wanted to do this, I think it would be possible via:

1. Add a script/`make` target, or similar, to launch `mkdocs build --strict`.
2. In that wrapper around `mkdocs build --strict`, check if the [Cloudflare Pages-specific environment variable](https://developers.cloudflare.com/pages/platform/build-configuration#environment-variables) — `CF_PAGES_BRANCH` — is `main`.
3. If the branch is `main` then set a `SITE_URL` (or similar) environment variable to `https://docs.opensafely.org`
4. In `mkdocs.yml`, specify `site_name: !ENV SITE_URL` — see the [mkdocs documentation](https://www.mkdocs.org/user-guide/configuration/#environment-variables). If the `SITE_URL` environment variable is not set, then restore the default.

Alternatively, if it's possible to set custom environment variables for testing and production within Cloudflare Pages, then that would remove the need for the wrapper: we'd just need the `!ENV SITE_URL` in `mkdocs.yml`.

madwort:
Nice!

PR #337, opened 2021-07-21
ubuntu installation instructions

Jongmassey:
Installation instructions based on my own experience of installation on a fairly fresh Ubuntu Linux installation.


bloodearnest (commented):
I am not sure we should be documenting adding external archives by default.

In production, we use the ubuntu shipped version of docker (docker.io, version 20.10 is in ubuntu 20.04), and everything works on that, we don't need any super modern docker features. 

So I'd recommend we document installing that instead. This would also simplify the instructions considerably, most of 1) would go away, just `apt install docker.io`

StevenMaude:
I'm overall :+1: for adding this, because there might be some people wanting to use Ubuntu.

Not sure if it might work as another tab in Getting Started instead of a separate page. There is a risk that this gets cluttered, against a risk of it being on a separate page and not discoverable. Maybe someone else has a stronger view on this.

NB as more of an FYI for anyone reading as opposed to a specific comment: you can also configure [rootless Docker](https://docs.docker.com/engine/security/rootless/). Not yet tried it to see if there are issues with that, particularly with respect to OpenSAFELY. That's possibly out of scope here because it's a more complicated installation process.

Jongmassey:
I was following the instructions on the docker website and got the impression that the ubuntu-shipped docker-io packages were old/deprecated. If they aren't/are adequate for running opensafely then I wholly agree that we should use the main ubuntu repos. 

As to where it should go - I imagine it's a fairly niche interest and so would probably be clutter on the main installation/getting started instructions.

bloodearnest:
> I was following the instructions on the docker website and got the impression that the ubuntu-shipped docker-io package was old/deprecated. If isn't/ is adequate for running opensafely then I wholly agree that we should use the main ubuntu repos.

Our production environment runs on the ubuntu packaged version, so we should recommend that for parity.

The ubuntu version version is typically ~6 months behind the docker published deb. I consider this a feature, for this use case, as it under goes a lot more testing. We really don't want to support opensafely linux users with docker's latest releases, stability is good.

PR #344, opened 2021-07-26
List what the ProjectCoordinator role allows

ghickman:


docs/permissions.md
```diff
@@ -9,3 +9,10 @@ Additionally, the OpenSAFELY framework requires a user to be a member of the `op
 While your repository is private, you may need to add new collaborators who are not named in your original application. Any new collaborators have the same information governance obligations as the original study authors, and must therefore sign the same access agreements; please send any such requests to [team@opensafely.org](mailto:team@opensafely.org).
 
 
+## Project Member Management
+
```
sebbacon, line 13: We could add some preamble here. How about something like this

"When they are approved to use the platform, every OpenSAFELY user is assigned roles that define what they are permitted to do. By default, most researchers using the platform will have `X` role, which permits them to X Y Z.  Additionally, users with the `ProjectCoordinator` role can.... <etc>.

sebbacon (changes requested):

ghickman:
@sebbacon – updated, how does that read?

docs/permissions.md
```diff
@@ -9,3 +9,18 @@ Additionally, the OpenSAFELY framework requires a user to be a member of the `op
...
+When they are approved to use the platform, every OpenSAFELY user is assigned roles that define what they are permitted to do.
+By default, most researchers using the platform will have the ProjectDeveloper role, which permits them to:
+
+* run jobs
+* cancel jobs
+* view job outputs
+* release reviewed outputs
```
sebbacon, line 20: Is this actually true right now? Not quite, right?

sebbacon: as in, I think it's necessary but not sufficient?

ghickman: hah, say you're on the outputs voyage without saying you're on the outputs voyage 😉

Removed until we have that process deployed fully.

sebbacon (approved):

PR #353, opened 2021-08-09
Reusable actions

iaindillingham:
This PR adds a section for reusable actions, moving the documentation from opensafely-core/actions-registry#7.

### DIFF
--- a/docs/actions-reusable.md
+++ b/docs/actions-reusable.md
@@ -1,0 +1,86 @@
+Like [scripted actions](actions-scripts.md), reusable actions are logical units of analytic code.
+However, whereas a scripted action is written to solve a problem for one study and must be copied-and-pasted to solve a similar problem for another study, a reusable action is written to solve a problem for several studies *without copying-and-pasting between them*.
+This makes reusable actions ideal for tasks that must be completed by several studies, such as producing cross-tabulations, summary statistics, or deciles charts.
+
+## Running reusable actions
+
+You can browse existing reusable actions at <https://actions.opensafely.org>.
+Although each is different, they have a common API.
+Consider the following extract from a study's *project.yaml*:
+
+```yaml
+actions:
+  generate_my_cohort:
+    run: cohortextractor:latest generate_cohort
+    outputs:
+      highly_sensitive:
+        cohort: output/input.csv
+
+  run_a_reusable_action:
+    # We will run version `v1.0.0` of the reusable action called `a_reusable_action`.
+    # The reusable action accepts an argument; in this case, a path to a file.
+    run: a_reusable_action:v1.0.0 output/input.csv
+    # The reusable action accepts a configuration option;
+    # in this case, an output format.
+    config:
+      output-format: PNG
+    needs: [generate_my_cohort]
+    outputs:
+      moderately_sensitive:
+        output: output/output_from_a_reusable_action.png
+```
+
+In the above extract, the `run` and `config` properties of `run_a_reusable_action` are the common API.
+The `run` property, which is required, describes how the reusable action is run.
+It comprises the name and the version of the reusable action, and, optionally, one or more arguments.
+The `config` property, which is optional, describes configuration options.
+
+## Developing reusable actions
+
+A reusable action is a public repo within the [`opensafely-actions`](https://github.com/opensafely-actions) organisation.
+It has a `main` branch, which is the release branch;
+versions are marked with [tags](http://git-scm.com/book/en/v2/Git-Basics-Tagging).
+In the above extract from a study's *project.yaml*, for example, we will run version `v1.0.0` of the reusable action called `a_reusable_action`.
+
+The repo has the following minimal structure:
+
+```sh
+.
+├── README.md
+└── action.yaml
+```
+
+*README.md* contains information about the reusable action, which is displayed by <https://actions.opensafely.org>.
+
+*action.yaml* contains a run command, which is composed of a runtime image (either `python`, `r`, or `stata-mp`) and an entry point.
+The entry point is a path to a script.
+For example:
+
+```yaml
+run: python:latest python action/cli.py
+```
+
+Where *action/cli.py* is:
+
+```python
+def main():
+  print("A reusable action")
+
+if __name__ == "__main__":
+  main()
+```
+
+When developing a reusable action, just as when developing a scripted action, the action's dependencies are made available by the runtime; they are not made available by the action.
+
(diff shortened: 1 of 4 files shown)


docs/actions-reusable.md
```diff
@@ -0,0 +1,63 @@
...
+  run_a_reusable_action:
+    # The reusable action is called "a_reusable_action".
+    # We will run version "v1.0.0".
+    run: a_reusable_action:v1.0.0
+    needs: [generate_my_cohort]
+    outputs:
+      moderately_sensitive:
+        output: output/output_from_a_reusable_action.csv
+```
+
```
sebbacon, line 24: Is it worth a bit more prose and/or examples and/or a link to safetab to explain how actions can also have configuration?

docs/actions-reusable.md
```diff
@@ -0,0 +1,63 @@
+Like [scripted actions](actions-scripts.md), reusable actions are logical units of analytic code.
+However, whereas the source code for a scripted action must be copied from study to study, the source code for a reusable action is shared by different studies.
+
+## Running reusable actions
+
+You can browse existing reusable actions at <https://actions.opensafely.org>.
+Although each is different, they have a common API.
```
inglesp, line 7: It's never spelt out what this API is.

docs/actions-reusable.md
```diff
@@ -0,0 +1,63 @@
...
+
+```yaml
+actions:
+  generate_my_cohort:
+    # ...
+
+  run_a_reusable_action:
+    # The reusable action is called "a_reusable_action".
+    # We will run version "v1.0.0".
+    run: a_reusable_action:v1.0.0
```
inglesp, line 18: Could this include an argument or two?  And maybe a demonstration of the `config:` stuff?

docs/actions-reusable.md
```diff
@@ -0,0 +1,63 @@
...
+```python
+def main():
+  print("A reusable action")
+
+if __name__ == "__main__":
+  main()
+```
+
+When developing a reusable action, just as when developing a scripted action, the action's dependencies are made available by the runtime; they are not made available by the action.
+In practice, this means that a Python action's *requirements.txt* is ignored.
```
inglesp, line 61: How does an action author know which dependencies are available?

docs/actions-reusable.md
iaindillingham: Eventually, yes 🙂 For now, I wanted to move the documentation from the issue to a more visible, permanent location. `safetab` also needs attention. Whilst it's working as a reusable action (opensafely-actions/test-reusable-actions#1), I'd prefer to improve its documentation and fix some inconsistencies first.

inglesp (approved):
Have suggested a couple of places where things could be clarified, but generally good stuff.

docs/actions-reusable.md
```diff
@@ -0,0 +1,63 @@
...
+    # We will run version "v1.0.0".
+    run: a_reusable_action:v1.0.0
+    needs: [generate_my_cohort]
+    outputs:
+      moderately_sensitive:
+        output: output/output_from_a_reusable_action.csv
+```
+
+## Developing reusable actions
+
```
sebbacon, line 26: Mention the template?

docs/actions-reusable.md
```diff
@@ -0,0 +1,63 @@
+Like [scripted actions](actions-scripts.md), reusable actions are logical units of analytic code.
+However, whereas the source code for a scripted action must be copied from study to study, the source code for a reusable action is shared by different studies.
```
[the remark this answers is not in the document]
iaindillingham: Do you mean that the distinction between scripted actions and reusable actions isn't helpful? Or that reusable actions have more advantages than the given advantage? (code reuse)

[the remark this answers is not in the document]
inglesp: Sorry, wasn't clear.

The distinction between scripted and reusable actions is certainly helpful!

But I don't think it's helpful to say

> whereas the source code for a scripted action must be copied from study to study

without referring to the reason you might do this.

docs/actions-reusable.md
iaindillingham: I've added an argument and a configuration option to the extract from a study's *project.yaml*. I'm keen to avoid using specific reusable actions as examples at this stage, because there's only `safetab` and `safetab` needs attention (see above).

docs/actions-reusable.md
iaindillingham: Good point 🙂 I've spelt out what we mean by "a reusable API".

docs/actions-reusable.md
iaindillingham: I've described where to find dependencies.

docs/actions-reusable.md
iaindillingham: Mentioned 🙂

docs/actions-reusable.md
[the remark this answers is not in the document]
StevenMaude: A slight rewording or change of focus in this sentence might help make clear the reasons for us having reusable actions.

My personal understanding is that:

* reusable actions aim to reduce researchers repeatedly solving the same problems by providing OpenSAFELY-related code in an easily accessed code library
* reusable actions can be used in a study, just by specifying them in `project.yaml`, without having to directly include the action's code in the study repository
  * using code that is available as a reusable action also reduces copy-pasting and maintenance of code between studies

(*The above isn't my suggested wording, just what I think the core ideas are.*)

Is it possible to describe the concept and motivation more directly in this opening?

docs/actions-reusable.md
StevenMaude: @iaindillingham: If this documentation page is the canonical reference for reusable actions, you might also want to link back on <https://actions.opensafely.org> to this page once it exists. You might have already thought of this, but just in case not :slightly_smiling_face:

docs/actions-reusable.md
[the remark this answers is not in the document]
iaindillingham: Super. I've attempted to clarify why code might be copied-and-pasted between studies: essentially to solve a similar problem in the second study as it did in the first study.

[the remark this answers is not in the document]
iaindillingham: Refreshed the page and saw your comment @StevenMaude. Does the update address it? I've emphasised the copying-and-pasting bit and the tasks that must be completed by several studies bit.

docs/actions-reusable.md
iaindillingham: Yup, that's on the list! 🙂

Issue #252, opened 2021-06-04
Study definition hacks


wjchulme:
# study definition haacks

This issue is for documenting some useful study definition workarounds that allow users to extract information in a way that is not formally built into the existing study definition API. 

They are collected here to help users do more with study definitions, and to help users and developers think about how the study definition API might be improved in future.

## Create a variable for each code in a codelist

Say we have a codelist and want to return the number of times each code in that codelist was used, then a variable is needed for each code within the codelist. The standard way to create these variables would be as follows:

```py
study = StudyDefinition(

    ...

    count_code1 = patients.with_these_clinical_events(
        codelist(code1, system="snomed"),
        on_or_after="index_date",
        returning = "number_of_matches_in_period",
        return_expectations={
            "incidence": 0.1,
            "int": {"distribution": "normal", "mean": 3, "stddev": 1}
        }
    ),

    count_code2 = patients.with_these_clinical_events(
        codelist(code2, system="snomed"),
        on_or_after="index_date",
        returning = "number_of_matches_in_period",
        return_expectations={
            "incidence": 0.1,
            "int": {"distribution": "normal", "mean": 3, "stddev": 1},
        }
    ),

    ...
	
)

```

A neater way of doing this is to wrap the variable signature in a function and loop that function over each code in the codelist:

```py

def loop_over_codes(code_list):

    def make_variable(code):
        return {
            f"count_{code}": (
                patients.with_these_clinical_events(
                    codelist([code], system="snomed"),
                    on_or_after="index_date",
                    returning="number_of_matches_in_period",
                    return_expectations={
                         "incidence": 0.1,
                         "int": {"distribution": "normal", "mean": 3, "stddev": 1},
                    },
                )
            )
        }

    variables = {}
    for code in code_list:
        variables.update(make_variable(code))
    return variables

```

This function is then invoked inside the `StudyDefinition` call for example as follows:

```py
study = StudyDefinition(

    ...
	
    **loop_over_codes(diabetes_codes),
	
    ...
)

```

This pattern can be used for any type of variable that uses codelists. For instance to extract the first admission date for a set of diagnoses in a codelist, use the following variable signature function:

```py
def make_variable(code):
    return {
        f"date_of_first_{code}": (
            patients.admitted_to_hospital(
                with_these_primary_diagnoses = codelist([code], system="icd10"),
                on_or_before="index_date",
		find_first_match_in_period=True,
                returning="date_admitted",
		return_expectations={
		    "incidence": 0.1,
[...]

PR #357, opened 2021-08-11
Add section on programming tricks for study definitions

wjchulme:
based on this issue https://github.com/opensafely/documentation/issues/252

### DIFF
--- a/docs/study-def-tricks.md
+++ b/docs/study-def-tricks.md
@@ -1,0 +1,249 @@
+This section introduces some tricks for working more efficiently with the study definition API, by helping you to avoid copy-and-pasting chunks of code unnecessarily. 
+
+## Create a variable for each code in a codelist
+
+Say we have a codelist and want to return the number of times each code in that codelist was used, then a variable is needed for each code within the codelist. The standard way to create these variables would be as follows:
+
+```py
+study = StudyDefinition(
+
+    ...
+
+    count_code1 = patients.with_these_clinical_events(
+        codelist(code1, system="snomed"),
+        on_or_after="index_date",
+        returning = "number_of_matches_in_period",
+        return_expectations={
+            "incidence": 0.1,
+            "int": {"distribution": "normal", "mean": 3, "stddev": 1}
+        }
+    ),
+
+    count_code2 = patients.with_these_clinical_events(
+        codelist(code2, system="snomed"),
+        on_or_after="index_date",
+        returning = "number_of_matches_in_period",
+        return_expectations={
+            "incidence": 0.1,
+            "int": {"distribution": "normal", "mean": 3, "stddev": 1},
+        }
+    ),
+
+    ...
+	
+)
+
+```
+
+A neater way of doing this is to wrap the variable signature in a function and loop that function over each code in the codelist:
+
+```py
+
+def loop_over_codes(code_list):
+
+    def make_variable(code):
+        return {
+            f"count_{code}": (
+                patients.with_these_clinical_events(
+                    codelist([code], system="snomed"),
+                    on_or_after="index_date",
+                    returning="number_of_matches_in_period",
+                    return_expectations={
+                         "incidence": 0.1,
+                         "int": {"distribution": "normal", "mean": 3, "stddev": 1},
+                    },
+                )
+            )
+        }
+
+    variables = {}
+    for code in code_list:
+        variables.update(make_variable(code))
+    return variables
+
+```
+
+This function is then invoked inside the `StudyDefinition` call for example as follows:
+
+```py
+study = StudyDefinition(
+
+    ...
+	
+    **loop_over_codes(diabetes_codes),
+	
+    ...
+)
+
+```
+
+This pattern can be used for any type of variable that uses codelists. For instance to extract the first admission date for a set of diagnoses in a codelist, use the following variable signature function:
+
+```py
+def make_variable(code):
+    return {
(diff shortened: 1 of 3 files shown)


docs/study-def-tricks.md
```diff
@@ -0,0 +1,318 @@
...
+		return_expectations={
+		    "incidence": 0.1,
+                    "date": {"2015-01-01": "index_date"}}
+                },
+            )
+        )
+    }
+```
+
+Example code: https://github.com/opensafely/long-covid/blob/master/analysis/study_definition_cohort.py
```
sebbacon, line 100: Make this a permalink?

docs/study-def-tricks.md
```diff
@@ -0,0 +1,318 @@
...
+        }
+    ),
+
+    ...
+	
+)
+
+```
+
+A neater way of doing this is to wrap the variable signature in a function and loop that function over each code in the codelist:
```
sebbacon, line 38: Personally I found the term _variable signature_ here a bit confusing (and more so when it reoccurs a few lines later); I wonder if it reads better just saying _variable_?

docs/study-def-tricks.md
```diff
@@ -0,0 +1,318 @@
...
+        }
+
+    variables = {}
+    for code in code_list:
+        variables.update(make_variable(code))
+    return variables
+
+```
+
+This function is then invoked inside the `StudyDefinition` call for example as follows:
```
sebbacon, line 66: A bit verbose? "This function can be invoked in the `StudyDefinition` section as follows:"?

docs/study-def-tricks.md
```diff
@@ -0,0 +1,318 @@
...
+	
+	...
+)
+```
+
+This can be used for any variable where you want to return information for a series of consecutive events. 
+
+The will create `n` columns in the dataset. A few things to be aware of:
+
+* This does not make for efficient SQL queries. It is creating a new query for each event, rather than a query for each set of events. This will slow down extraction time considerably.
```
sebbacon, line 188: "This does not make for efficient SQL queries" <- the user doesn't necessarily know the relationship between study definitions and SQL. Is there a way to reword that's simply about performance in general?

docs/study-def-tricks.md
```diff
@@ -0,0 +1,318 @@
...
+* The dummy data won't necessarily appear in date order. It may be possible to tinker with the function to make this work. Another workaround is to reorder the columns post-extraction, for example here: https://github.com/opensafely/covid-vaccine-effectiveness-research/blob/be747894e1fadf525e391c01477a8ac532613b42/analysis/R/data_process.R#L229 
+* It's likely that you'll want to to convert this wide-formatted data to long format,
```
sebbacon, line 190: Trailing comma, was there something you wanted to add?

docs/study-def-tricks.md
```diff
@@ -0,0 +1,318 @@
...
+        },
+    ),
+    
+    ....
+
+)
+
+```
+
+If the category levels are named with a predictable structure for example (`["stage 1", "stage 2", ...]`, or [`100, 200, 300, ...`]) then a dict comprehension can be used to create the levels, without having to import them. For example for IMD ranks rounded to the nearest 100:
```
sebbacon, line 233: link to dict comprehensions in python docs?

docs/study-def-tricks.md
```diff
@@ -0,0 +1,318 @@
+This section introduces some tricks for working more efficiently with the study definition API, by helping you to avoid copy-and-pasting chunks of code unnecessarily.
```
sebbacon, line 1: I think we need to add a few words to warn that these are all python examples and will probably be hard to adapt for someone without a bit of python experience?

I think we could add something similar in this part in the main Study Def page, and link to this "hacks" page:

> It is written in the Python programming language, using an OpenSAFELY-specific format which is intended to be easily written, read, and reviewed by anyone with some epidemiological knowledge.

Something like "when you find you are copy and pasting a lot, you may find it easier to write loops in python code, which does require a little more python skill - examples here"...?

sebbacon: We should also invite people who come up with their own hacks to share them with us! It's a great way of learning common patterns that we should consider integrating into the framework.

docs/study-def-tricks.md
alexwalkercebm: +1

alexwalkercebm (approved):
I agree with Seb's comments. One extra one.

docs/study-def-tricks.md
wjchulme: Sticking with official python documentation, I could link to [dictionaries](https://docs.python.org/3/tutorial/datastructures.html#dictionaries) or [list comprehensions](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions), but not dict comprehensions. So I've linked to both -- hopefully that's not a terrible idea

docs/study-def-tricks.md
wjchulme: Thanks, I've reworded

wjchulme:
@sebbacon sebbacon and @alexwalkercebm I just copy-pasted this from the issue but should've properly scrutinised first. Thanks for the suggestions

StevenMaude:
@wjchulme: :+1: for adding this. It is far more likely that study authors will see these ideas in the documentation, than they would in a less discoverable issue.

PR #360, opened 2021-08-11
Describe Outputs Roles

ghickman:
This adds what the various outputs roles allow users to do.

I tried explaining these in prose based on the releasing processes (view on L4, release to job-server, create draft snapshot, publish snapshot), but it didn't make much sense on the permissions page.


ghickman:
@StevenMaude – good shout, done and rebased in.

PR #370, opened 2021-08-17
Explain more on Gitpod usage and requirements

StevenMaude:
Clarify a couple of things relating to Gitpod.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -29,13 +29,11 @@ In this guide, we've documented two different ways to work with OpenSAFELY:
    installed for you*.
 
      Services such as [Gitpod](https://gitpod.io) provide online
-     computing environments; Gitpod currently has a free plan with a
-     generous monthly usage limit for working with public code. We've
-     adapted our demonstration research study and this guide such that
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -49,6 +47,20 @@ In this guide, we've documented two different ways to work with OpenSAFELY:
...
+       Edge, Firefox, Safari);
+     * your network connection does not block access to the Gitpod
+       development environment.
+
+     If your internet connection is not managed by you directly —
+     perhaps you are in the office or connecting via your employer's
+     virtual private network (VPN) — it may be that [corporate firewalls
+     prevent access to
+     Gitpod](https://github.com/opensafely/documentation/discussions/311);
+     please contact the IT staff who manage the network for help.
```
madwort, line 62: I think `the network` is slightly ambigious here - how about the same phrase used earlier in the paragraph (i.e. `your internet connection` )? Possibly poor style in conventional writing, but I think better in this context?

madwort (changes requested):
nitpick: ambiguity

madwort (approved):

PR #372, opened 2021-08-18
Describe New Releasing Process

### LINKED ISSUE #371: Reference Current State of Releasing Output
Releasing of outputs to the jobs site is currently rolling out, the docs should 1) explain that! and 2) link to the various [internal] resources for staff so they can move users onto the new process and know what permissions are required.

Relatedly the various releasing terms should be glossaryified.

ghickman:
This adds some context on the new release process and links the glossary to the permissions page so we can define some terms for releasing outputs.

Fixes #371

### DIFF
--- a/docs/permissions.md
+++ b/docs/permissions.md
@@ -50,3 +50,5 @@ For releasing outputs we have some additional roles:
 **OutputPublisher:**
 
 * Confirm publication of outputs
+
+---8<-- 'includes/glossary.md'
--- a/docs/permissions.md
+++ b/docs/permissions.md
@@ -29,6 +29,12 @@ Additionally, users with the **ProjectCoordinator** role can:
 
 ## Releasing Outputs
 
+!!! note
+    Releasing of outputs to the job server is currently being rolled out.
+    See [team manual](https://bennettinstitute-team-manual.pages.dev/tech-team/releasing-outputs/) for instructions on setting up a user.
+    The new release process is described separately in [Google Docs](https://docs.google.com/document/d/1PSzMoCFov2olJpxrFGT1DccWUqGDA67mUvPIyIhOcew/edit#heading=h.aokkzsfuxo4n).
+
+
 For releasing outputs we have some additional roles:
 
 **ProjectCollaborator:**
--- a/docs/permissions.md
+++ b/docs/permissions.md
@@ -31,7 +31,7 @@ Additionally, users with the **ProjectCoordinator** role can:
 
 !!! note
     Releasing of outputs to the job server is currently being rolled out.
-    See [team manual](https://bennettinstitute-team-manual.pages.dev/tech-team/releasing-outputs/) for instructions on setting up a user.
+    See [team manual](https://bennettinstitute-team-manual.pages.dev/tech-team/playbooks/job-server-releasing-outputs/) for instructions on setting up a user.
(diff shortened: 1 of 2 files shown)


docs/permissions.md
```diff
@@ -29,6 +29,11 @@ Additionally, users with the **ProjectCoordinator** role can:
 
 ## Releasing Outputs
 
+!!! note
+    Releasing of outputs to the Jobs site is currently being rolled out.
```
StevenMaude, line 33: Should "Jobs site" be "OpenSAFELY Jobs site"? Not sure, but maybe. I'm not sure if we're consistent throughout the documentation.

Not a big problem though.

docs/permissions.md
```diff
@@ -29,6 +29,11 @@ Additionally, users with the **ProjectCoordinator** role can:
...
+    Setting up a user is documented [here](https://bennettinstitute-team-manual.pages.dev/tech-team/releasing-outputs/) and the new process is documented [here](https://docs.google.com/document/d/1PSzMoCFov2olJpxrFGT1DccWUqGDA67mUvPIyIhOcew/edit#heading=h.aokkzsfuxo4n).
```
StevenMaude, line 34: Maybe avoid "here" links if possible?

includes/glossary.md
```diff
@@ -33,3 +33,5 @@
 *[CLI]: Command-line interface, a text interface used by some computer software that is usually run via a "terminal" or "command-line shell".
 *[2FA]: Two-factor authentication, a means of verifying a user is authorised to access a service by checking for something that only the user should know (e.g. a password) and something the user should have (e.g. an authenticator application installed on the user's phone).
 *[API]: Application Programming Interface. An agreed way for one system to communicate with another system.
+*[released outputs]: files from Level 4 uploaded to the Jobs site.
+*[draft published outputs]: a collection of outputs prepared for publication to the general public.
```
sebbacon, line 37: a collection of _released outputs_?

docs/permissions.md
ghickman: This is the first use of that term, previously we've used `job server`, I'll stick with that to be consistent for now and make a note to use a new term when we rename it.

docs/permissions.md
ghickman: Are they considered not ideal? Not sure what to replace them with, any suggestions would be greatly appreciated!

StevenMaude: Maybe something like:

> See the [team manual]() for instructions on setting up a user. The new release process is described separately in [Google Docs](https://docs.google.com/document/d/1PSzMoCFov2olJpxrFGT1DccWUqGDA67mUvPIyIhOcew/edit#heading=h.aokkzsfuxo4n).

NB: I don't know if there is a definitive singular noun equivalent to "Google Docs": "Google Doc"?

***

As for motivation, this is very much opinion-based over evidence, but suggestions in [this Stack Exchange question](https://ux.stackexchange.com/questions/12100/why-shouldnt-we-use-words-such-as-here-and-this-in-textlinks), that "here" links:

* are less helpful for screen reader users skimming links; I found a post on [WebAIM](https://webaim.org/techniques/hypertext/#link_to_link) that backs this up.
* are slightly misguiding in drawing the reader's eye to text with little useful meaning
* contain an extra word that is usually redundant

ghickman: Thank you on both fronts!  Updated text as suggested and rebased in.

sebbacon: Another reason is for SEO (not that it matters in this case, but a good habit to get into)

StevenMaude (approved):

PR #383, opened 2021-08-31
Add gzip instructions to cohortextractor action

HelenCEBM:
Instruct users to use csv.gz format for cohortextractor


evansd (approved):
Sorry! I missed the review request and just noticed this.

Thanks very much for doing this.

PR #392, opened 2021-09-03
Restructure menu

HelenCEBM:
Should create a more logical structure with fewer high-level menu items, hopefully easier to navigate. We should ensure that the "best practices" sections are referenced elsewhere to help users find them. Some sections/intros might need rewording to fit with the new structure.


wjchulme (approved):
This is much better. I've made a couple of further changes too.

annaschultze (approved):
This looks really good - much more intuitive to navigate I think. Thanks both!

PR #398, opened 2021-09-09
update cohort-extractor to pull in number_of_matches_in_period in sgss

CarolineMorton:
update cohort-extractor to 1.44.1


StevenMaude:
@CarolineMorton: It probably would be better if cohort-extractor is automatically bumped quickly after a new release.

In the meantime, you can run the [Dependabot check manually](https://github.com/opensafely/documentation/network/updates) if you need to do this [1]: otherwise, it runs daily. I've just done this to create #399.

That also has the advantage that you can review the PR yourself :heavy_check_mark:

[1] Assuming there aren't already [five other Dependabot pull requests](https://github.com/opensafely/documentation/blob/353f8339813a48d638c7001652caaa79832ab9b3/.github/dependabot.yml#L7) that aren't cohort-extractor open.

CarolineMorton:
Awesome @StevenMaude thanks for this! I was going around in circles. I followed the [INSTALL.md](https://github.com/opensafely/documentation/blob/main/INSTALL.md) page but as you can see ended up in a pickle... (btw should we delete that page ...)

Anyway thank you !

PR #402, opened 2021-09-10
Update measures.md

HelenCEBM:
Make it explicit that index_date is needed, and also add notes on minimising run time when extracting multiple cohorts.

### DIFF
--- a/docs/measures.md
+++ b/docs/measures.md
@@ -83,8 +83,12 @@ measures = [
 ]
 ```
 
-This differs from a normal study definition due to the addition of the `measures` object, which is a list of calls to the `Measure()` function, for each measure.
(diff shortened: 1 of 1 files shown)


wjchulme (approved):
I've made a change to highlight the issue with having to define `index_date` even iff it will be overwritten by `index-date-range`.

I'm also not sure if the "Index dates" sub-section is needed? It's covered later on in the "extract the data" section. Maybe some of that info could be moved. But if it stays as is it's fine!

HelenCEBM:
> I've made a change to highlight the issue with having to define index_date even iff it will be overwritten by index-date-range.

I wrote this in as well but deleted it again!

PR #404, opened 2021-09-10
Clarify instructions on updating cohort-extractor

StevenMaude:
There was some confusion in #398 regarding updating. This is primarily
due to me not updating the instructions here. This commit aims to reduce
that confusion.


CarolineMorton (approved):
Awesome thanks for updating this!

PR #407, opened 2021-09-15
Dates in between clause

HelenCEBM:
Couple of notes about dates being in the right order when supplied to a 'between' clause. 

Also separately amended the variables page.

### DIFF
--- a/docs/study-def-dates.md
+++ b/docs/study-def-dates.md
@@ -5,8 +5,9 @@ In study definitions, dates are described in `"YYYY-MM-DD"` format; so, for exam
 
 Most variable extractor functions have arguments for specifying the date range over which you want to retrieve information.
 Most commonly this is `on_or_before=`, `on_or_after=`, or `between=` (see the [variable reference](study-def-variables.md) for full documentation).
-You must use at most one.
+You must use at most one. 
 If no option is given then it will use all dates (including possibly future dates).
+
 
 As well as specifying dates explicitly with e.g., `on_or_before="2019-12-31"`, you can use the date expressions discussed [below](#index-dates).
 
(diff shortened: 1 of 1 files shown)


docs/study-def-dates.md
```diff
@@ -5,9 +5,10 @@ In study definitions, dates are described in `"YYYY-MM-DD"` format; so, for exam
 
 Most variable extractor functions have arguments for specifying the date range over which you want to retrieve information.
 Most commonly this is `on_or_before=`, `on_or_after=`, or `between=` (see the [variable reference](study-def-variables.md) for full documentation).
-You must use at most one.
+You must use at most one.
```
acagreen17, line 8:
```suggestion
At most, you can use one of these functions.
```

docs/study-def-dates.md
```diff
@@ -40,7 +41,7 @@ study = StudyDefinition(
 )
 ```
 
-This can make it easier to change the index date of a study by making sure it is only defined in one place.
+This can make it easier to change the index date of a study by making sure it is only defined in one place. (Note that if using `between`, ensure that the dates are given in chronological order).
```
acagreen17, line 44: add example? e.g., `between = "index_date - 1 month", "index_date`

acagreen17 (approved):
Looks good. Add a couple of suggestions, but feel free to ignore.

docs/study-def-dates.md
HelenCEBM: Thinking about it I don't actually think this it's true that you can't use more than one - you can use both `on_or_before=` and `on_or_after=`, you just can't combine them with `between`. Is that right?

HelenCEBM: HAve made a further change to this - how does it look?

docs/study-def-dates.md
HelenCEBM: I've referred to the example that is directly above since it is already there, does that look ok?

acagreen17: Looks good

acagreen17:
All good to merge I think 👍

PR #421, opened 2021-09-24
More on csv.gz format

HelenCEBM:
Make it more clear that csv.gz format is recommended but csv may need to be used instead if Stata is being used for analysis

### DIFF
--- a/docs/actions-cohortextractor.md
+++ b/docs/actions-cohortextractor.md
@@ -56,15 +56,15 @@ A basic `generate_cohort` action looks like this:
 
 ```yaml
 generate_study_cohort
-  run: cohortextractor:latest generate_cohort --output-format =csv.gz
+  run: cohortextractor:latest generate_cohort --output-format=csv.gz
   outputs:
     highly_sensitive:
       data: output/input.csv.gz
-    ### This file is produced in compressed format. If using Stata for your analysis you may need to remove 
-    ### the --output-format option and the .gz file extension to produce an uncompressed CSV instead.
+    ### This file is produced in compressed format. If using Stata for your analysis you may need to change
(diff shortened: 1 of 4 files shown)


docs/actions-cohortextractor.md
```diff
@@ -56,13 +56,15 @@ A basic `generate_cohort` action looks like this:
 
 ```yaml
 generate_study_cohort
-  run: cohortextractor:latest generate_cohort --output-format=csv.gz
+  run: cohortextractor:latest generate_cohort --output-format =csv.gz
```
evansd, line 59: I think you can either use a space or an equals here, but I'm not sure about both. I think it should be either:
```
--output-format=csv.gz
```
or
```
--output-format csv.gz
```

docs/actions-cohortextractor.md
```diff
@@ -56,13 +56,15 @@ A basic `generate_cohort` action looks like this:
...
   outputs:
     highly_sensitive:
       data: output/input.csv.gz
+    ### This file is produced in compressed format. If using Stata for your analysis you may need to remove 
+    ### the --output-format option and the .gz file extension to produce an uncompressed CSV instead.
```
evansd, line 64: Would it be clearer to say "use `--output-format=csv`", rather than to say "remove the option"? No particular strong feeling, but it seems a bit more consistent to me.

evansd (approved):
Thanks @HelenCEBM, this looks good. A couple of small comments but otherwise good to merge I think.

docs/actions-cohortextractor.md
HelenCEBM: Oops thanks

docs/actions-cohortextractor.md
HelenCEBM: OK done

PR #427, opened 2021-09-28
Mention merging the releases branch

HelenCEBM:
Mention explicitly that release-candidates branch should be merged into main just to make sure users know what to do with it.

### DIFF
--- a/docs/releasing-files.md
+++ b/docs/releasing-files.md
@@ -15,6 +15,8 @@ Level 4 outputs **can only be released by specific members of the OpenSAFELY tea
 
 !!! note
     Individual researchers who have Level 4 access have responsibility for redacting sensitive information or choosing not to publish it at all. The study author should do everything they can to make this easy; for example, carrying out low number suppression automatically, documenting code clearly, and only selecting essential items for publication when deciding what to label as `moderately_sensitive`.
+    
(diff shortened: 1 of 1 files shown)


sebbacon (approved):

wjchulme:
This will be fine for most people. But...

If the workspace is based on a branch other than `main`, say `devbranch`, then you will probably want to merge with `devbranch` instead of `main`, as the outputs really belong on the same branch as the code that created them. Merging to `devbranch` is made slightly more complicated by the fact that the `release-candidates` branch that's created when you release the files is based on `main` (probably -- see below), so there may be conflicts when you try to merge to with `devbranch`.

The branch that `release-candidates` is based on is determined as follows:
* if `release-candidates` doesn't already exist on the remote, then it is based on `main`. This is the usual situation, since `release-candidates` is automatically deleted once it has been merged
* if `release-candidates` does exist on the remote, then the new releases are added to this existing `release-candidates` branch. This would occur if:
    * The `release-candidates` branch hasn't yet been merged
    * The `release-candidates` branch was created _by the user_, in order to base it on a branch other than `main` (for example with `git checkout -b release-candidates devbranch`). This option is useful to avoid the conflicts mentioned above.

This detail may be too complicated for the documentation. But it's worth writing down in case we need to point users to this in future. Hopefully not for long once the output viewer is in full swing!

HelenCEBM:
How about "... should be merged into **its parent branch (usually `main`)** ..."

PR #477, opened 2021-11-03
feat: add outpatient appointment study def var

madwort:
* I'd like cohortextractor to be bumped to `1.50.0` before merging this


StevenMaude:
Looks good to me now, following #478.

StevenMaude (approved):

madwort:
yep looks good

PR #482, opened 2021-11-09
Warn ICNARC data is for ICNARC researchers only

iaindillingham:
Thank you @amirmehrkar for providing the text of this warning.

I've searched the repo for instances of ICNARC. These two warnings cover them, with the exception of this incidental mention, which is on the [SystmOne primary care](https://docs.opensafely.org/dataset-systmone/) page:

> For some datasets (ICNARC), OpenSAFELY receives the hashed NHS numbers from the external dataset and the matching occurs inside OpenSAFELY. The matched IDs are then sent back and the matched records are returned to OpenSAFELY.

I do not think we need a warning on this page.

Closes opensafely-core/cohort-extractor#678

### DIFF
--- a/docs/dataset-icnarc.md
+++ b/docs/dataset-icnarc.md
@@ -1,6 +1,10 @@
 The ICNARC-CMP  dataset contains information on covid-related intensive care admissions in England.
 
 !!! warning
+    ICNARC data can only be used in collaboration with ICNARC researchers who must be involved in working on the study and writing it up.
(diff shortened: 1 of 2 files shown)


amirmehrkar:
Thanks @iaindillingham I agree not needed here.

I will also update the policy pages I'm writing to mention the entry gateway too.

amirmehrkar (approved):
Thanks for updating content re ICNARC collaboration

PR #483, opened 2021-11-09
Ethnicity study def tips

LFISHER7:
This adds some instructions for extracting ethnicity when you are running your study definition over multiple time points to the study definition tips section.

@milanwiedemann this should hopefully clear up your questions from last week, but let me know if it isn't clear


docs/study-def-tricks.md
```diff
@@ -319,4 +319,63 @@ study = StudyDefinition(
     ),
     **common_variables
 )
-```
\ No newline at end of file
+```
+
+## Extracting ethnicity across multiple time points
+
+Querying ethnicity can take a long time as a result of having to map from individual codes to broad categories. This is ok if you only extract ethnicity at one time point, but if you are running your study definition across multiple time points it can slow down the cohort extraction. As ethnicity is generally expected to remain constant over time, you can instead extract ethnicity once (taking the latest recorded ethnicity code for each patient) and then map these to the files generated by your main study definition for each time point.
```
milanwiedemann, line 326: are there good examples for  "individual codes" and "broad categories" that we could add here?

Querying ethnicity can take a long time as a result of having to map from individual codes (e.g., XX) to broad categories (e.g., YY).

docs/study-def-tricks.md
```diff
@@ -319,4 +319,63 @@ study = StudyDefinition(
...
+        default_expectations={
+            "date": {"earliest": "1900-01-01", "latest": "today"},
+            "rate": "uniform",
+        },
+
+        population=patients.all(),
+
+        # ETHNICITY IN 6 CATEGORIES
+        eth=patients.with_these_clinical_events(
+            ethnicity_codes,
```
milanwiedemann, line 342: add comment here saying where these `ethnicity_codes` come from or what's in this codelist? or can we expect that this is known if people look at this section?

milanwiedemann (commented):
thanks, I think this is good! see what you think about my two small comments, feel free to ignore. was thinking if we should link to the sro-template analysis files (or any other public repo that does this) at the end https://github.com/opensafely/sro-template/tree/master/analysis?

docs/study-def-tricks.md
LFISHER7: I'll add this in

docs/study-def-tricks.md
LFISHER7: I think this probably deserves its own section somewhere, defining ethnicity is a bit of a can of worms in itself. What to use for EMIS, adding in hospital recorded ethnicity etc. I'll make a note to put it somewhere else

LFISHER7:
The above adds a link to the sro-template repo as an example use case and points to the ethnicity codelist where codes are mapped to broader categories

milanwiedemann (approved):
sry, missed this! looks great

PR #484, opened 2021-11-10
Clarify `--output-format` argument

iaindillingham:
Clarify that the `--output-format` argument is passed to `generate_cohort` and not `generate_measures`. The latter will automatically work with compressed output formats, but will always output CSV files.

### DIFF
--- a/docs/measures.md
+++ b/docs/measures.md
@@ -182,6 +182,18 @@ However the final step, which combines output across time periods, will always b
 This command accepts an `--output-dir` argument, which defines the directory in which the output measure files will be created. 
 If the `--output-dir` argument is configured, `generate_measures` expects the input cohort files to also be in the output directory. To ensure this, include the same `--output-dir` argument in the previous `generate_cohort` step.
 
+!!! note
+    As we saw, you can extract the data in a compressed output format, such as `csv.gz`, `feather`, `dta`, or `dta.gz`.
+    For example:
+
+    ```sh
+    cohortextractor generate_cohort --output-format=csv.gz
+    ```
+    
+    If you do, then the `generate_measures` command will automatically work with this compressed output format;
+    you need not pass `--output-format` to `generate_measures`, in other words.
+    However, the `generate_measures` command will always output CSV files.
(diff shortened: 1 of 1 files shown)


docs/measures.md
```diff
@@ -86,7 +86,7 @@ measures = [
 This differs from a normal study definition due to the addition of the `measures` object, which is a list of calls to the `Measure()` function, for each measure. 
 
 #### The `index_date`
-Each month/week start date covered by your study period is passed to the study definition as the `index_date` in turn, thus allowing the cohort and variables to vary each month/week as required. Some variables can retain fixed or no date ranges, e.g. a person's ethnicity would not be expected to vary over time. Read more about using index_date [here](/study-def-dates.md/#index-dates).
+Each month/week start date covered by your study period is passed to the study definition as the `index_date` in turn, thus allowing the cohort and variables to vary each month/week as required. Some variables can retain fixed or no date ranges, e.g. a person's ethnicity would not be expected to vary over time. Read more about using `index_date` [here](/study-def-dates.md/#index-dates).
```
StevenMaude, line 89: nit, and only because it's caught my eye now: If we're tweaking this line anyway, could we rewrite this slightly to remove the use of "here" for the link.

Not essential, but would be nice :sparkles:

See [a previous discussion ~~here~~](https://github.com/opensafely/documentation/pull/372#discussion_r692155882).

StevenMaude (approved):

docs/measures.md
iaindillingham: Nice strike-through 😆 

Thank you for drawing my attention to the link. I'm actually going to remove it, because:
* It's broken. This is probably a formatting issue, because [the target](https://docs.opensafely.org/study-def-dates/#index-dates) still exists. More importantly, though...
* The information contained at the target isn't applicable. You can't supply a date expression when generating measures (i.e. with `--index-date-range`) and the `index_date` in the study definition is overridden. It seems unreasonable to say "See this thing with the same name, that doesn't work in the same way" 🙂

PR #505, opened 2021-12-02
Backend Contracts Docs

ghickman:
This installs our mkdocs plugin and adds the backend contracts page.

I've included the JSON file from cohort-extractor-v2 until we have a way to make generating it part of the build process.


ghickman:
Blocked by opensafely-core/mkdocs-opensafely-backend-contracts#2

INSTALL.md
```diff
@@ -21,6 +21,18 @@ If you have added new functions, you will need to add an explicit
 reference to them in
 [docs/study-def-variables.md](./docs/study-def-variables.md).
 
+
+### Updating Cohort Extractor v2 Definitions
+
+cohort-extractor-v2 is a documentation dependency which we can't install directly.
```
StevenMaude, line 27: suggestion: Maybe a one sentence explanation or link to relevant issue is useful here? I assume it's because of the legacy Python version on Cloudflare Pages.

INSTALL.md
```diff
@@ -21,6 +21,18 @@ If you have added new functions, you will need to add an explicit
...
+Instead it generates a JSON file which we commit to this repo.
+
+Run `python -m cohortextractor2 extract-contract-docs` to generate the JSON file.
```
StevenMaude, line 30: thought: Maybe we could later add a GitHub Action in this repository to generate the file, and create a PR if the file changes?

docs/backend-contracts.md
```diff
@@ -0,0 +1,19 @@
+# Table-centric documentation
```
StevenMaude, line 1: suggestion (non-blocking): This page will be visible to someone browsing the docs. So, perhaps adding an initial, maybe almost a placeholder, section that indicates:

* who this documentation page is aimed at
* what the page describes
* that this documentation page will be frequently changed

Something like (rough pass, wording could be improved; I'm not sure if we're capitalising "Backend Contracts" everywhere or not):

```markdown
# Introduction to Backend Contracts

This documentation on OpenSAFELY's Backend Contracts is a reference for data providers who wish to integrate their backend systems into OpenSAFELY.

Backend Contracts are an OpenSAFELY specification. By satisfying some or all of these Contracts, backend providers can certify their datasets for suitability with OpenSAFELY.

!!! warning

    This specification is currently under rapid development and is subject to frequent change.
```

Doesn't necessarily have to be in this PR though.

mkdocs.yml
```diff
@@ -57,6 +57,8 @@ nav:
       - Developing a protocol: protocol.md
       - Using Git effectively: git-workflow.md
       - Code reviews: code-reviews.md
+  - Building for OpenSAFELY:
```
StevenMaude, line 60: nit (non-blocking): I feel like it's possible to have a slightly more precise title here. Unhelpfully, I don't know what might be. "Building datasets for OpenSAFELY?" "Integrating with OpenSAFELY"?

It's fine to leave for now in this PR and change later.

INSTALL.md
ghickman: I suspect that's the way we'll go, yep.  v2 isn't quite there with its versioning story but I expect we'll do some kind of version check too.

mkdocs.yml
ghickman: Yeah, I think this is a task for later, this PR is just about getting the backend docstrings in.

StevenMaude (approved):
summary: I've left some optional suggestions, but looks good overall.

It would be better to not have to manually build the data. But fine for now given current constraints of building this documentation.

docs/backend-contracts.md
StevenMaude: Yep, fine!

StevenMaude: Oh, just to mention, because I wasn't sure and tested. It's possible to have the new `.md` file unlisted in the table of contents, but still be included in the build, if you know the URL. (That works even with `mkdocs build --strict`, which surprised me a bit.)

Maybe useful in future, if you want to have it previewed on the main docs site for people, but not yet "public".

PR #511, opened 2021-12-08
fix: install git alongside github desktop

Jongmassey:
Installation of GitHub desktop does not place `git` on the path.
This breaks some functionality of the `opensafely-cli`

### DIFF
--- a/docs/install-github-and-git.md
+++ b/docs/install-github-and-git.md
@@ -13,14 +13,13 @@ To get set up, you will need:
 
 
 ## New to git
-For Windows or macOS users new to git, the easiest way to do both is to install [**GitHub Desktop**](https://desktop.github.com/).
-This includes a command line version of git and a useful GUI for editing files and gitting.
+For Windows or macOS users new to git, [**GitHub Desktop**](https://desktop.github.com/) provides a useful GUI for editing files and gitting.
 To install GitHub Desktop, visit the [GitHub Desktop homepage](https://desktop.github.com/) and click install for your operating system.
 Once installed, it will ask you to create or sign-in to your GitHub account.
 Follow the instructions as appropriate.
 
 Once signed in, you'll be taken to a _Let's get started_ page.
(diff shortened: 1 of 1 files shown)


docs/install-github-and-git.md
```diff
@@ -13,14 +13,13 @@ To get set up, you will need:
...
-For Windows or macOS users new to git, the easiest way to do both is to install [**GitHub Desktop**](https://desktop.github.com/).
-This includes a command line version of git and a useful GUI for editing files and gitting.
+For Windows or macOS users new to git, [**GitHub Desktop**](https://desktop.github.com/) provides a useful GUI for editing files and gitting.
 To install GitHub Desktop, visit the [GitHub Desktop homepage](https://desktop.github.com/) and click install for your operating system.
 Once installed, it will ask you to create or sign-in to your GitHub account.
 Follow the instructions as appropriate.
 
 Once signed in, you'll be taken to a _Let's get started_ page.
-You'll need to install other OpenSAFELY requirements before diving in.
+You'll need to install other OpenSAFELY requirements before diving in, including the [command line version of git](https://git-scm.com/).
```
iaindillingham, line 22: I was going to ask whether GH desktop has an option to install `git`. It does, apparently, have [a CLI](https://docs.github.com/en/desktop/installing-and-configuring-github-desktop/overview/launching-github-desktop-from-the-command-line). Just not the CLI we'd like 🤷🏻‍♂️.

iaindillingham (approved):

docs/install-github-and-git.md
evansd, line 22: These feels a bit like burying the lede to me. Would it be better as something like?

> Note, you'll still need to install the command line version of git alongside Github Desktop, as well as the other OpenSAFELY requirements, before diving in.

evansd (approved):

docs/install-github-and-git.md
Jongmassey: Good shout

PR #513, opened 2021-12-09
Mention that Apple-based Macs can't access Level 4

StevenMaude:
At least, right now.

This commit aims to save other users from spending time trying this and
discovering it doesn't work the long way round.

### DIFF
--- a/docs/releasing-files.md
+++ b/docs/releasing-files.md
@@ -5,6 +5,31 @@ The final tier is review of so-called "Level 4" outputs, where the OpenSAFELY fr
 If you have Level 4 access you can log in and review your aggregated results, and perform any necessary redactions before requesting files to be released. 
 You will find your files to review in `D:\Level4Files\workspaces\<workspacename>`.
 
+!!! note
+
+    **Mac users**
+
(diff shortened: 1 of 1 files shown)


docs/releasing-files.md
```diff
@@ -5,6 +5,31 @@ The final tier is review of so-called "Level 4" outputs, where the OpenSAFELY fr
...
+
+    However, Mac users with newer Macs that have *Apple* processors —
+    for example, the M1 processor — can only run Windows in a virtual
+    machine via the Windows on ARM release: **this configuration is
+    currently incompatible with the client necessary for Level 4
+    access.**
+
+    Macs with Apple processors are still suitable for writing, testing
+    and submitting OpenSAFELY code to be run on the secure server.
+    **This issue only affects access to Level 4 server.**
```
StevenMaude, line 27: I assume that this is true. Confirmation from a real Mac M1 user would be good.

tomodwyer: Confirmed this is working locally on MacBook Air (M1, 2020)

tomodwyer (approved):

PR #528, opened 2022-01-05
Update opensafely-cohort-extractor to 1.52.2

StevenMaude:
Manually for now, because Dependabot is not working correctly for some
reason. Dependabot drops other dependencies: for example, #524.

This is a recent problem. For instance, #508 opened by Dependabot did
actually correctly upgrade the package to 1.51.5.

### DIFF
--- a/requirements.prod.txt
+++ b/requirements.prod.txt
@@ -441,9 +441,9 @@ numpy==1.21.4 \
     #   pyarrow
     #   scipy
     #   seaborn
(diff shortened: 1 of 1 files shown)


StevenMaude:
Instead of correcting the existing Dependabot pull request (#524), I've left that pull request intact but closed it.

This is so that I can cite it "cleanly" in a bug report to the Dependabot issue tracker.

iaindillingham (approved):

PR #542, opened 2022-01-18
Fix default PYTHON_VERSION in justfile

iaindillingham:
The default Python version should be consistent between:

* `justfile`
* `.python-version`
* `.github/workflows/build.yml`

---

Follow up: As per @benbc's [suggestion][1], it would be good to add an assertion to `build.yaml` to check for consistency between `justfile` and `.python-version`. I think this would require refactoring `build.yml`, which seems non-trivial:

* Checking that `actions/setup-python` doesn't set `PYTHON_VERSION`
* Calling `just prodenv` from within `build.yml` (rather than `pip install ...`)
* Testing that `python --version` is equal to `cat .python-version`, or nearly equal (not necessarily to the patch release)

Could I double-check my understanding of ☝🏻? If I'm on the right track, then I'll make a separate issue.

[1]: https://ebmdatalab.slack.com/archives/C63UXGB8E/p1642520205077800

### DIFF
--- a/justfile
+++ b/justfile
@@ -23,7 +23,7 @@ clean:
 virtualenv:
     #!/usr/bin/env bash
     # allow users to specify python version in .env
-    PYTHON_VERSION=${PYTHON_VERSION:-python3.9}
+    PYTHON_VERSION=${PYTHON_VERSION:-python3.7}
 
     # create venv and upgrade pip
     test -d $VIRTUAL_ENV || { $PYTHON_VERSION -m venv $VIRTUAL_ENV && $PIP install --upgrade pip; }


ghickman (approved):

ghickman:
I'm not sure syncing the justfile, `.python-version` and `setup-python` versions is worth the effort.  AIUI this is the first time we've hit this edge case, so I'd argue we need to hit it a few times more before we consider scripting it away (ie prove it's actually an issue worth dev time).

An easier fix would be wrapping `mkdocs build --strict` from `build.yaml` into a justfile command and calling that so we get `prodenv` as a dependency of that command.

benbc:
@iaindillingham I'm not familiar with this codebase, so my ideas may not be helpful. I had imagined a `Justfile` target (called by `check`, maybe?) which asserts that whatever's in the file matches whatever's in the `Justfile`.

iaindillingham:
Thanks for your comments, @ghickman and @benbc. I'll pause the follow-up; if the inconsistency trips me up again, then I'll revisit.

PR #543, opened 2022-01-19
Document running Jupyter notebooks via `opensafely`

StevenMaude:
Document that you can run JupyterLab via the `opensafely` command-line tool.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -773,6 +773,8 @@ against real data.  [Read more about permissions here](permissions.md).
 In the meantime, take a look at the rest of our documentation for more
 detail on the subjects covered in this tutorial. For example:
 
+* There is a [more complete guide](opensafely-cli.md) to the OpenSAFELY
+  command-line tool.
 * The [full study definition reference](study-def.md) describes all the
   different ways to define new variables in your study definition.
 * You'll find more information about the contents of `project.yaml` in the
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -773,8 +773,8 @@ against real data.  [Read more about permissions here](permissions.md).
 In the meantime, take a look at the rest of our documentation for more
 detail on the subjects covered in this tutorial. For example:
 
-* There is a [more complete guide](opensafely-cli.md) to the OpenSAFELY
-  command-line tool.
+* There is a more complete [guide to the OpenSAFELY command-line
+  tool](opensafely-cli.md).
 * The [full study definition reference](study-def.md) describes all the
   different ways to define new variables in your study definition.
 * You'll find more information about the contents of `project.yaml` in the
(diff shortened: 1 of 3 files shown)


docs/opensafely-cli.md
```diff
@@ -125,3 +125,21 @@ To pull the most recent Docker images to your machine, run:
...
+environments for developing code.
+
+You can run JupyterLab to use Jupyter notebooks via the `opensafely`
+tool. This ensures that the Python code you write will work in the
+OpenSAFELY environment.
+
+To use Jupyter notebooks via `opensafely`:
+
+1. Run `opensafely jupyter` to start a JupyterLab server on your
+   computer.
```
bloodearnest, line 141: add "in the directory containing the code you would like to run with jupyter" ?

docs/opensafely-cli.md
```diff
@@ -125,3 +125,21 @@ To pull the most recent Docker images to your machine, run:
...
+2. You will see long URLs starting `http://localhost…` and
```
bloodearnest, line 142: It should automatically open a brower

bloodearnest (changes requested):

docs/getting-started.md
```diff
@@ -773,6 +773,8 @@ against real data.  [Read more about permissions here](permissions.md).
 In the meantime, take a look at the rest of our documentation for more
 detail on the subjects covered in this tutorial. For example:
 
+* There is a [more complete guide](opensafely-cli.md) to the OpenSAFELY
```
madwort, line 776: as part of the accessibility guidelines, I think we need to make sure we're using meaningful link text - I don't think `[more complete guide] to the cli` stands on its own well enough as link text, how about `more complete [guide to the cli]`?

madwort (changes requested):
nitpick link text, otherwise looks good

docs/opensafely-cli.md
StevenMaude: I'll mention that as the default, as it seems like it should be the case.

(This isn't the behaviour for me, and @iaindillingham said [similar](https://github.com/opensafely/documentation/discussions/530). This is why I specified the copy-paste step explicitly.)

docs/getting-started.md
StevenMaude: Yep, agree!

bloodearnest:
Added a minor suggestion.

I don't want to land this until I've finished the testing I'm doing with internal users, as there's still some issues for some folks.

It's very difficult to test this feature, so my plan was to release quietly, having trusted users test on their machines, and then announcing in docs once it's more solid.

madwort (approved):
wfm

StevenMaude:
@bloodearnest: sounds good!

bloodearnest (approved):

bloodearnest:
Right, the command seems to be broadly working, so good to go from my pov

PR #550, opened 2022-01-24
Use pyenv's new :latest suffix to get latest minor version

ghickman:
Pyenv has [added a :latest suffix to the install command](https://github.com/pyenv/pyenv/pull/1831), allowing one to do something like `pyenv install 3.10:latest` to get the latest 3.10 your pyenv knows about (homebrew installed pyenv is often one or two patch releases of Python behind, but I don't think that affects us).

This updates the docs to make use of that suffix, also bumping the Python version we install to 3.10.

Note: I've not gone with `3:latest` because that picks up the 3.11 alpha.

### DIFF
--- a/docs/install-macos.md
+++ b/docs/install-macos.md
@@ -45,19 +45,22 @@ Use pyenv to install Python:
 !!! note "This command might take a while to run depending on the speed of your computer."
 
 ```bash
-pyenv install 3.9.1
+pyenv install 3.10:latest
 ```
 
-Then enable version 3.9.1 in pyenv:
+Look for the line `Installing Python-3.10.n...` (where `n` is a number).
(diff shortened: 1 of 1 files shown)


ghickman:
@iaindillingham – we had 3.9 in here previously, and I've just gone with "latest minor version" every time I've touched this document!  `opensafely-cli` works on 3.10 so I think it should be fine.  Do you have links to the opensafely-cli specific bits which talk about 3.8? I can only see references to 3.8 in a commented out line of the Python installation docs.

ghickman:
@iaindillingham – with you!  opensafely-cli has a minimum version of 3.8, which the README.md reflects, so it's just DEVELOPERS.md which needs an update.  We can probably remove the version from there completely, I'm not sure what benefit it brings there.

PR #555, opened 2022-01-25
Rename Support Pages to Match their Titles

ghickman:
This hopefully makes them easier to find when working on a local checkout.


docs/case-control-studies.md
```diff
@@ -102,7 +102,7 @@ actions:
     [python-docker](https://github.com/opensafely-core/python-docker),
     [r-docker](https://github.com/opensafely-core/r-docker), or
     [stata-docker](https://github.com/opensafely-core/stata-docker) repositories.
-    Alternatively, you can [request new packages](requests-packages.md).
+    Alternatively, you can [request new packages](requesting-libraries.md).
```
StevenMaude, line 105: suggestion: Maybe should be

```suggestion
    Alternatively, you can [request new libraries](requesting-libraries.md).
```

StevenMaude (approved):

PR #556, opened 2022-01-25
Set up Toolchain for Including Databuilder Code Example

ghickman:
This sets up a toolchain for writing examples in Python modules, testing them, breaking them down into snippets, and generating markdown files for inclusion into docs pages.

The general flow is:
* write python modules in `databuilder/snippets/`
* wrap sections of the python modules in [snex](https://github.com/jwbargsten/snex) markers
* extract snippets to markdown files in `examples/`
* include the snippets in the same way we include the glossary

The tooling for this is all handled inside the `databuilder` directory, including requirements, dedicated justfile, separate virtualenv, etc so that we can install the databuilder project into a virtualenv using a valid version of Python (it's minimum is 3.9).

There is our standard stack of Python tooling (black, flake8, isort) set up for working on the modules themselves, and the modules are directly test by running them with `python <module>`.  I've also set up a watch command for snex and configured mkdocs to watch the `examples` directory so pages update when changes are made to the example modules.

I tried out `doctest` but it wasn't a particularly nice flow for our use case, we lose the dev tooling and don't really gain anything over just executing the examples.

This toolchain pushes us to write examples in the documentation repo against a pinned version of databuilder on the basis that quick feedback of how examples look in the generated documentation is more ideal than tying examples to a released version of databuilder.


docs/updating-the-docs.md
```diff
@@ -30,3 +30,21 @@ README](https://github.com/opensafely/documentation#building-locally-and-testing
 that details the use of `pip-compile` for this.
 
 ---8<-- 'includes/glossary.md'
+
+## Making changes to the data definition examples
```
StevenMaude, line 34: note: I started rewriting this section slightly in the `steve/databuilder-examples` branch. There's just an extra commit there.

databuilder/justfile
```diff
@@ -0,0 +1,118 @@
+# make a new virtualenv so we don't use the main docs one, which is why we're
+# not using $VIRTUAL_ENV here.
+export VIRTUAL_ENV  := ".venv"
+
+# TODO: make it /scripts on windows?
```
StevenMaude, line 5: question: What about running any of this on Windows?

Again, not absolutely necessary now, but if we want non-developer contributors to add to this, then it might be something to think about.

An alternative might be to have a `workflow_dispatch` workflow that builds the snippets for you, (and maybe checks them first) and avoids a local checkout.

databuilder/justfile
```diff
@@ -0,0 +1,118 @@
...
+
+# clean up temporary files
+clean:
+    rm -rf .venv
+
+
+# ensure valid virtualenv
+virtualenv:
+    #!/usr/bin/env bash
+    PYTHON_VERSION=$(which python3.10)
```
StevenMaude, line 25: note: This caught me out because I don't have Python 3.10 installed. But not a big deal.

databuilder/snippets/dsl.py
```diff
@@ -0,0 +1,24 @@
+# :snippet imports
```
StevenMaude, line 1: thought: I guess that it would be possible to separate out examples into subdirectories if need be (because for instance, we might want a self-contained project with `project.yaml` or there are conflicting behaviours between different modules?) 

I've not tested this though.

docs/data-definition.md
```diff
@@ -0,0 +1,75 @@
+!!! danger "This page is for use with [the OpenSAFELY Databuilder](https://github.com/opensafely-core/databuilder/) which is currently unreleased."
```
StevenMaude, line 1: thought: I haven't extensively read through this or know it's exactly what we want, but it's fine as a starting point, given this isn't "live" documentation.

databuilder/justfile
ghickman: The justfile stack doesn't support windows currently because of this line.  Simon and I have discussed it a few times and we think it should be simple to support with a condition on the OS, we've just not had the need yet (but assume it will come eventually).

Building examples with a workflow should be trivial, the workflow in this PR already does that, but it doesn't commit the change.  The only downside to consider with that is seeing the outcome of those changes in the docs would mean changes need to go through the workflow first.

examples/src-data-definition.md
```diff
@@ -0,0 +1,17 @@
+```python
```
StevenMaude, line 1: question: Material for MkDocs has a few nice features for [fenced code blocks](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#code-blocks). These require specifying configuration alongside the opening backticks.

I'm not sure if we can get this working with snex. There is some option of templating, but I don't think that's flexible enough to let us customise that some code blocks might have `\`\`\` py title="dsl.py"` and others `\`\`\`py hl_lines="4 5"`.

You can customise the language in the `# :snippet…` lines, but you can't use quotes as they get escaped.

If there's no workaround, is a better approach is to let the author add their own backticks? It's slightly more work for the author, but then allows for all the flexibility that Material for MkDocs offers.

databuilder/snippets/dsl.py
ghickman: Not sure I follow the example but breaking them out into subdirectories could be tricky, snex has some hard code behaviour on name mapping, however it's an open source tool so we could potentially fork if it was necessary.

examples/src-data-definition.md
StevenMaude: There's a similar question on still supporting [code annotations](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#code-annotations). It's easiest to see how this works in the [raw Markdown](https://raw.githubusercontent.com/squidfunk/mkdocs-material/b2ecd717bdebc71b514d3eb0ace10efc497fcf98/docs/reference/code-blocks.md) for those docs; see under "Code annotations".

In principle, this is simpler, because you could have a separate snippet, maybe inside a docstring inside the Python, that points to the annotation. But can we mark a snippet as "we don't want backticks". This also suggests having the author write backticks might be more flexible overall.

StevenMaude:
summary: Overall, this is a good approach. I have a few questions about the flexibility of how this would work: that is, does everything Material for MkDocs support still work? I think there would be workarounds for these.

The fact that it's not yet polished for non-developers is less of an issue right now, and we could think about improving that later.

examples/src-data-definition.md
ghickman: You can change the output template for the files it produces, see the `output_template` config line here: https://github.com/jwbargsten/snex#advanced-snippet-tagging

StevenMaude (approved):

PR #569, opened 2022-02-01
Set up pre-commit hooks for the databuilder directory

ghickman:
This sets up pre-commit for only the databuilder directory, using the global `files` key.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,36 @@
+default_language_version:
+  python: python3.10
+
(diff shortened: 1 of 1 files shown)


StevenMaude:
There's still a hardcoded Python version here, as in the Justfile, but that's fine for now.

StevenMaude:
For anyone else reading, this is so we have some basic commit-time checking of Data Builder code examples to be included in documentation.

StevenMaude (approved):

PR #571, opened 2022-02-02
Add Script to Update Databuilder Version

ghickman:
It seems like dependabot won't touch the version of databuilder, which is logical since it's pinned and thus pip-compile won't touch it either.

So this adds a script to bump the version up when needed.


databuilder/update-version.py
```diff
@@ -0,0 +1,43 @@
+import re
+from pathlib import Path
+
+import requests
+
+
+version_pat = re.compile(r"\/v.*\.zip")
```
StevenMaude, line 7: nit: This regex could be more selective to pick out the `databuilder zip`. As it stands, if we have another `.zip` in there, we'll overwrite it. Though it's unlikely we'll have more than one `.zip` in the requirements.

StevenMaude (approved):

databuilder/update-version.py
ghickman: I had a more selective one in at first, but with `.sub()` call you have to replace the matched string, not just a capture group, so a more selective pattern has to duplicate the URL format where it's called.  Given how unlikely we are to have other zips I just kept it to the simple pattern!

PR #578, opened 2022-02-07
therapeutics dataset documentation

rebkwok:


docs/dataset-therapeutics.md
```diff
@@ -0,0 +1,34 @@
...
+
+## More Information
+
+...Links to any other sources
+
+## Datasource-specific glossary
+
+...Any specific terms to define
+* **Term** description
+Each treatment is a row.  Patients should only appear once per episode treated, but may legitimately appear more than once if they received a second treatment for a separate episode.
```
rebkwok, line 31: This is almost certainly not right, but we probably want to describe what multiple rows per patient should mean?

HelenCEBM: Each row is one completed form for one course of treatment. The patient may or may not have actually received the treatment or completed the course (but we assume that they usually do). They may have another form completed for another treatment, either because it was decided to give them a different treatment, or for some other reason. They may in theory also have another form completed some months later for another instance of infection.

docs/dataset-therapeutics.md
```diff
@@ -0,0 +1,34 @@
+The Therapeutics dataset contains information on ...
+
+!!! warning
+    Therapeutics data can only be used ...
```
brianmackenna, line 4: @amirmehrkar can you remind me who this is restricted to please?

HelenCEBM:
@brianmackenna

acagreen17:
Would it be useful to define the possible treatment types and risk groups, like we do with status and indications?

docs/dataset-therapeutics.md
HelenCEBM: @amirmehrkar has confirmed it is just subject to the usual restrictions, nothing additional

HelenCEBM (approved):

PR #601, opened 2022-02-18
Document background to OpenSAFELY Contracts

StevenMaude:
This PR covers an initial pass at explaining OpenSAFELY Contracts.

The actual reference documentation will be included separately; that
documentation will be generated automatically.

It additionally adds the beginnings of a page aimed at backend data
providers. This additional page also discusses Contracts.

### DIFF
--- a/docs/contracts-intro.md
+++ b/docs/contracts-intro.md
@@ -2,11 +2,181 @@
 
 ---8<-- 'includes/data-builder-danger-header.md'
 
-This page will introduce OpenSAFELY Contracts, explaining:
+## OpenSAFELY Contracts are a clinical data specification
 
-* what they are
-* the process by which new contracts could be proposed
-* where [details of existing contracts will be
-  found](contracts-reference.md)
+*OpenSAFELY Contracts* are an important OpenSAFELY concept that is
+introduced with Data Builder.
+
+This page describes:
+
+* what OpenSAFELY Contracts are
+* how OpenSAFELY Contracts relate to dataset definitions
+
+Each OpenSAFELY Contract is a formal specification, [described in
+code](https://github.com/opensafely-core/databuilder/blob/main/databuilder/contracts/contracts.py),
+relating to a particular clinical data domain. OpenSAFELY Contracts are
+authored by the OpenSAFELY platform developers: the Contract designs are
+informed by discussion and feedback from data providers and researchers.
+
+Researchers write dataset definitions to access data tables from
+OpenSAFELY backends. All available data tables within OpenSAFELY satisfy
+an OpenSAFELY Contract, regardless of which data provider operates the
+backend.
+
+TODO: does a table only satisfy one contract, or possibly multiple?
+
+## The OpenSAFELY Contracts reference
+
+The [OpenSAFELY Contracts reference](contracts-reference.md) has full
+details of each individual Contract. When writing a dataset definition,
+researchers should consult [the OpenSAFELY Contracts
+reference](contracts-reference.md) to see:
+
+* the data available in OpenSAFELY backends
+* how the data is structured
+* what that data represents
+
+!!! warning
+    OpenSAFELY Contracts apply only when accessing OpenSAFELY data using
+    *dataset definitions*, written to run with Data Builder. Contracts
+    do not apply to *study definitions* written for the existing cohort
+    extractor.
+
+## The goals of OpenSAFELY Contracts
+
+Contracts make it easier for:
+
+* *Researchers* to understand what data is available in
+  OpenSAFELY and how that data is represented.
+* *Data providers* to structure their data for OpenSAFELY.
+* *OpenSAFELY's developers* to extend OpenSAFELY to additional
+  healthcare and clinical data domains in future.
+
+!!! note
+    OpenSAFELY Contracts are a new concept still being explored.
+
+    We [value
+    feedback](https://github.com/opensafely/documentation/discussions)
+    from both OpenSAFELY users and data providers, and conversations on
+    how we might further refine the design of individual OpenSAFELY
+    Contracts or the system as a whole.
+
+### The benefits of OpenSAFELY Contracts for researchers
+
+With OpenSAFELY Contracts, researchers benefit from:
+
+* Guarantees on the structure of the data table being queried.
+* A simpler query interface to multiple backends with the ability to:
+    * look up which backends provide the data of interest
+    * verify which backends are compatible with their dataset definitions
+    * run data definitions on multiple different backends, without
+      having to write specific code for each individual backend
+* More consistent documentation that:
+    * describes the commonalities of the data offered by different backends
+    * notes where individual backends provide data that deviates from others
+
+## Common and Custom Contracts
+
+Data backends that are accessible via OpenSAFELY provide data tables. Each
+data table satisfies an OpenSAFELY Contract.
+
+There are two types of Contracts:
+
+* **Common Contracts**, required by all OpenSAFELY backends and covering
+  core clinical data domains
+* **Custom Contracts**, offered by just some backends and covering more
+  specialised clinical data domains
+
+Both the structure of the Contracts and the method of accessing
+associated tables is the same for both Common and Core Contracts.
+
+### Common Contracts
+
+All OpenSAFELY backends satisfy Common Contracts.
+
+There will be several different Common Contracts available with the
+initial release of Data Builder. More Common Contracts may be added in
+future.
+
+!!! warning
+    Some backends might deviate slightly from the suggested OpenSAFELY
+    behaviour shared by most backends. Refer to the [OpenSAFELY
+    Contracts Reference](contracts-reference.md) for backends you intend
+    to access with your dataset definition.
+
+For dataset definitions that only access tables defined by Common
+Contracts, those dataset definitions should be largely compatible with all
+OpenSAFELY backends.
+
+### Custom Contracts
+
+Some backends may offer particularly specialised data that is not yet
+covered by a Common Contract. *Custom Contracts* allow individual
+backends to still offer this data via OpenSAFELY.
+
+!!! warning
+    Using Custom Contracts might tie a dataset definition more closely
+    to a specific backend, than if just Common Contracts are used. This
+    also depends on how the dataset definition is designed.
+
+!!! note
+    Should additional data providers want to offer data that closely resembles
+    an existing *Custom* Contract, OpenSAFELY developers would consider a new
+    *Common* Contract.
+
+## Reading an OpenSAFELY Contract
+
+!!! note
+    Refer to the [OpenSAFELY Contracts
+    Reference](contracts-intro#the-opensafely-contracts-reference) for
+    the details of specific contracts.
+
+Each OpenSAFELY Contract relates to a specific domain of health record
+data. The name of the Contract indicates the data domain covered by the
+Contract. Tables that satisfy these Contracts can be accessed in a
+dataset definition via [Data Builder's query language,
+ehrQL](ehrql-intro.md).
+
+An OpenSAFELY Contract provides the following information:
+
+* the **variable names**. Each variable is denoted by the
+  identifier you use to access the variable in ehrQL.
+* the **variable's description**. The description summarises what the
+  variable represents.
+* the **variable's data type**. For instance, a variable
+  might be one of the following:
+    * integer type
+    * floating point type
+    * date type
+    * string type
+    * or some other type not in this list
+* the **additional constraints on the variable's data values**, other
+  than those given by the variable's type. For instance,
+    * the variable might be of date type, but all values are the first
+      of the month
+    * the variable might be of integer type, but all values are positive
+      integers
+
+## Miscellaneous TODOs
+
+* TODO: Should we refer to columns or variables?
+* TODO: Finalise details above based on what we actually display. Do we
+  know all the variable types, or can we link to them?
+* TODO: Do we want to have nulls as a constraint or as a separate
+  column in the reference?
+* TODO: How are deviations noted in the reference? Is this via a note
+  within the same contract? Or are we going to document the same
+  contract multiple times for different providers.
+* TODO: Where do we document about how to validate which backends run a
+  definition?
+* TODO: Do we need anything on Contract versioning right now? How does
+  versioning work?
+
+## For backend data providers
+
+* TODO: Does content for backend providers go here or another page?
+* TODO: What is the process by which new contracts could be proposed?
+* TODO: Where do we describe the integration process for
+  data providers?
 
 ---8<-- 'includes/glossary.md'
--- a/docs/contracts-intro.md
+++ b/docs/contracts-intro.md
@@ -4,47 +4,52 @@
 
 ## OpenSAFELY Contracts are a clinical data specification
 
-*OpenSAFELY Contracts* are an important OpenSAFELY concept that is
-introduced with Data Builder.
+An *OpenSAFELY Contract* is a core concept in the OpenSAFELY framework.
+Anyone using Data Builder should familiarise themselves with how
+Contracts work.
 
 This page describes:
 
 * what OpenSAFELY Contracts are
 * how OpenSAFELY Contracts relate to dataset definitions
 
-Each OpenSAFELY Contract is a formal specification — [described in
-code](https://github.com/opensafely-core/databuilder/blob/main/databuilder/contracts/)
-— relating to a particular clinical data domain.
+Each OpenSAFELY Contract is a formal specification relating to a
+particular clinical data domain.
+
+!!! note
+    For researchers, the human-readable version of each Contract is
+    listed in the [Contracts reference](contracts-reference.md).
+
+    For developers working on OpenSAFELY integrations, you can also see
+    how [each Contract is specified in Python
+    code](https://github.com/opensafely-core/databuilder/blob/main/databuilder/contracts/).
+    **It is not necessary to understand the Contract's Python code to be
+    able to use a Contract.**
 
 OpenSAFELY Contracts are authored by the OpenSAFELY platform developers:
 the Contract designs are informed by discussion and feedback from [data
 providers](data-provider-integration.md) and researchers.
 
-Researchers write dataset definitions to access data tables from
-OpenSAFELY backends. All of the data tables available from OpenSAFELY
-satisfy a single OpenSAFELY Contract, regardless of which data provider
-operates the backend.
+### How do OpenSAFELY Contracts fit into OpenSAFELY?
 
-## The OpenSAFELY Contracts reference
+Researchers write [dataset definitions](dataset-definition.md) to access
+data tables from OpenSAFELY backends. This is a simple example of a
+dataset definition that accesses a data table:
 
-The [OpenSAFELY Contracts reference](contracts-reference.md) has full
-details of each individual Contract. When writing a dataset definition,
-researchers should consult [the OpenSAFELY Contracts
-reference](contracts-reference.md) to see:
+---8<-- 'examples/src-minimal-ehrql.md'
 
-* what data is available in OpenSAFELY backends
-* how the data is structured
-* what that data represents and how it should be interpreted
+Dataset definitions are run using OpenSAFELY Data Builder. When running
+the dataset definition against live backends on the OpenSAFELY platform,
+the data requested will then be extracted.
 
-!!! warning
-    OpenSAFELY Contracts apply only when accessing OpenSAFELY data using
-    *dataset definitions*, written to run with Data Builder. Contracts
-    do not apply to legacy [*study definitions*](study-def.md) written for
-    cohort-extractor.
+!!! note
+    The dataset definition is provided as an example only here.
+    You do not need to understand how to write a dataset definition
+    to follow the rest of this page.
 
-!!! warning
-    The OpenSAFELY Contracts are still in development and subject to
-    frequent change.
+**Data tables in OpenSAFELY backends must relate to an OpenSAFELY
+Contract.** Researchers then have guarantees that data from different
+data providers can be accessed and interpreted in a consistent way.
 
 ## The goals of OpenSAFELY Contracts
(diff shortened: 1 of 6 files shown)


StevenMaude:
As the commit message suggests, there are some TODOs.

Some of these may not be fixable right now, because maybe we don't know how to resolve them.

If anyone has opinions on what should be there in place of certain TODOs, I'm happy to add things in this PR.

docs/contracts-intro.md
```diff
@@ -2,11 +2,181 @@
...
+*OpenSAFELY Contracts* are an important OpenSAFELY concept that is
+introduced with Data Builder.
+
+This page describes:
+
+* what OpenSAFELY Contracts are
+* how OpenSAFELY Contracts relate to dataset definitions
+
+Each OpenSAFELY Contract is a formal specification, [described in
+code](https://github.com/opensafely-core/databuilder/blob/main/databuilder/contracts/contracts.py),
```
benbc, line 16: In fact that module is obsolete. The real contracts will be namespaced, for example the `universal` module. I suggest that you link to https://github.com/opensafely-core/databuilder/blob/main/databuilder/contracts which will have several modules and subfolders in it.

docs/contracts-intro.md
```diff
@@ -2,11 +2,181 @@
...
+relating to a particular clinical data domain. OpenSAFELY Contracts are
+authored by the OpenSAFELY platform developers: the Contract designs are
+informed by discussion and feedback from data providers and researchers.
+
+Researchers write dataset definitions to access data tables from
+OpenSAFELY backends. All available data tables within OpenSAFELY satisfy
+an OpenSAFELY Contract, regardless of which data provider operates the
+backend.
+
+TODO: does a table only satisfy one contract, or possibly multiple?
```
benbc, line 26: Only one at the moment and I think that a one-to-one relationship makes sense since the marginal cost of new tables is tiny.

docs/contracts-intro.md
```diff
@@ -2,11 +2,181 @@
...
+reference](contracts-reference.md) to see:
+
+* the data available in OpenSAFELY backends
+* how the data is structured
+* what that data represents
+
+!!! warning
+    OpenSAFELY Contracts apply only when accessing OpenSAFELY data using
+    *dataset definitions*, written to run with Data Builder. Contracts
+    do not apply to *study definitions* written for the existing cohort
```
benbc, line 42: "Existing" seems odd here, since the Data Builder will exist too by the time users read this.

docs/contracts-intro.md
```diff
@@ -2,11 +2,181 @@
...
+
+Data backends that are accessible via OpenSAFELY provide data tables. Each
+data table satisfies an OpenSAFELY Contract.
+
+There are two types of Contracts:
+
+* **Common Contracts**, required by all OpenSAFELY backends and covering
+  core clinical data domains
+* **Custom Contracts**, offered by just some backends and covering more
+  specialised clinical data domains
```
benbc, line 88: Is this correct. I'm not sure that we require _all_ backends to implement _all_ common contracts.

I thought the distinction was that common contracts are defined by us and if you want to implement them you have to do so completely and meet all the constraints; whereas custom contracts are defined by a backend provider to cover some data or format specific to their backend.

docs/contracts-intro.md
```diff
@@ -2,11 +2,181 @@
...
+* the **additional constraints on the variable's data values**, other
+  than those given by the variable's type. For instance,
+    * the variable might be of date type, but all values are the first
+      of the month
+    * the variable might be of integer type, but all values are positive
+      integers
+
+## Miscellaneous TODOs
+
+* TODO: Should we refer to columns or variables?
```
benbc, line 162: I think these are columns. We use "variable" for a different thing -- an element of a dataset definition.

docs/contracts-intro.md
```diff
@@ -2,11 +2,181 @@
...
+* TODO: Finalise details above based on what we actually display. Do we
+  know all the variable types, or can we link to them?
```
benbc, line 164: We will need to document all of the variable types, so it should be possible to link to them. I suppose we could also dynamically generate a complete list, but I'm not sure there's much value in being sure we're complete here.

docs/contracts-intro.md
```diff
@@ -2,11 +2,181 @@
...
+* TODO: Do we want to have nulls as a constraint or as a separate
+  column in the reference?
```
benbc, line 166: This is really out of scope for this page, but I think we should have columns for very common things like this (which I would call "required" rather than referring to nulls);

docs/contracts-intro.md
```diff
@@ -2,11 +2,181 @@
...
+!!! note
+    Should additional data providers want to offer data that closely resembles
+    an existing *Custom* Contract, OpenSAFELY developers would consider a new
+    *Common* Contract.
+
+## Reading an OpenSAFELY Contract
+
+!!! note
+    Refer to the [OpenSAFELY Contracts
+    Reference](contracts-intro#the-opensafely-contracts-reference) for
```
benbc, line 131: This link is broken.

docs/contracts-intro.md
```diff
@@ -2,11 +2,181 @@
...
+* TODO: Should we refer to columns or variables?
+* TODO: Finalise details above based on what we actually display. Do we
+  know all the variable types, or can we link to them?
+* TODO: Do we want to have nulls as a constraint or as a separate
+  column in the reference?
+* TODO: How are deviations noted in the reference? Is this via a note
+  within the same contract? Or are we going to document the same
+  contract multiple times for different providers.
+* TODO: Where do we document about how to validate which backends run a
+  definition?
```
benbc, line 171: We don't have this as a feature yet. I think we can worry about how to document it when we implement it.

docs/contracts-intro.md
```diff
@@ -2,11 +2,181 @@
...
+* TODO: Do we need anything on Contract versioning right now? How does
+  versioning work?
```
benbc, line 173: We don't know yet, but we have some ideas. Are you happy to defer this to later when we do the relevant coding work?

benbc (commented):
I've reviewed this fairly closely, but I'd really like to get @sebbacon's eyes on it as well.

docs/contracts-intro.md
StevenMaude: Yes, I think you're right, and this is my misreading.

docs/contracts-intro.md
StevenMaude: It's fine to leave this out and introduce it at the time when versioning is implemented.

docs/data-provider-integration.md
```diff
@@ -0,0 +1,83 @@
...
+researchers using OpenSAFELY to run studies across multiple data
+backends. However, there may be administrative or technical reasons why
+data providers cannot satisfy a Contract precisely.
+
+Some deviation from the Contract is permitted.
+
+TODO: What needs to be done if there is deviation? Is this following
+discussion and approval by OpenSAFELY?
+TODO: What counts as permitted deviation and what counts as a breach of
+the Contract?
```
StevenMaude, line 51: Alternate wording or clarification on this welcome. It's also OK to remove this section entirely for now, if we don't have this worked out yet.

sebbacon: We don't know yet. For me, this is part of the point, rather than just some handwaving to be resolved in a month or two.

My hypothesis is that the process of agreeing on a contract implementation will often be a conversation, or even a negotiation, between the data provider, and the opensafely contract owner(s). This negotiation will over time cause us (the contract owners) to iterate our understanding of what constitutes a particular domain, and what concepts can reasonably be applied across different implementations of that domain.

So there will never really be a final version, although I'd expect core concepts / contracts to asymptotically approach a final version over time.

Does that make sense? I think it's probably worth attempting to capture some of the flavour of this in these docs.

docs/contracts-intro.md
```diff
@@ -2,11 +2,159 @@
...
 
-This page will introduce OpenSAFELY Contracts, explaining:
+## OpenSAFELY Contracts are a clinical data specification
 
-* what they are
-* the process by which new contracts could be proposed
-* where [details of existing contracts will be
-  found](contracts-reference.md)
+*OpenSAFELY Contracts* are an important OpenSAFELY concept that is
+introduced with Data Builder.
```
sebbacon, line 8: I find this formulation slightly awkward. Partly because of the slippage from plural to singular, but mainly because of "introduced with".

Maybe something like:

> An OpenSAFELY "Contract" is a core concept in the OpenSAFELY framework. Anyone who uses Data Builder should spend time familiarising themselves with how they work.

?

I'm not normally picky but I think the opening sentence is super important!

docs/contracts-intro.md
```diff
@@ -2,11 +2,159 @@
...
+* how OpenSAFELY Contracts relate to dataset definitions
+
+Each OpenSAFELY Contract is a formal specification — [described in
+code](https://github.com/opensafely-core/databuilder/blob/main/databuilder/contracts/)
+— relating to a particular clinical data domain. OpenSAFELY Contracts
+are authored by the OpenSAFELY platform developers: the Contract designs
+are informed by discussion and feedback from [data
+providers](data-provider-integration.md) and researchers.
+
+Researchers write dataset definitions to access data tables from
```
sebbacon, line 22: I think an example here could be helpful.

Is this any clearer?

> When a researcher writes a dataset definition, they routinely access data tables from OpenSAFELY backends with code that looks like this:
> (example)
> Data tables must always relate to a single OpenSAFELY Contract; this provides a guarantee that researchers can access and interpret the data in a consistent way, regardless of which data provider supplies the underlying data.

docs/contracts-intro.md
```diff
@@ -2,11 +2,159 @@
...
+operates the backend.
+
+## The OpenSAFELY Contracts reference
+
+The [OpenSAFELY Contracts reference](contracts-reference.md) has full
+details of each individual Contract. When writing a dataset definition,
+researchers should consult [the OpenSAFELY Contracts
+reference](contracts-reference.md) to see:
+
+* what data is available in OpenSAFELY backends
```
sebbacon, line 34: what data is available in _which_ OpenSAFELY backends

?

docs/contracts-intro.md
```diff
@@ -2,11 +2,159 @@
...
+* how the data is structured
+* what that data represents
```
sebbacon, line 36:
```suggestion
* what that data represents and how it should be interpreted
```

docs/contracts-intro.md
```diff
@@ -2,11 +2,159 @@
...
+
+!!! warning
+    OpenSAFELY Contracts apply only when accessing OpenSAFELY data using
+    *dataset definitions*, written to run with Data Builder. Contracts
+    do not apply to [*study definitions*](study-def.md) written for
```
sebbacon, line 41:
```suggestion
    do not apply to legacy [*study definitions*](study-def.md) written for
```

docs/contracts-intro.md
```diff
@@ -2,11 +2,159 @@
...
+details of each individual Contract. When writing a dataset definition,
+researchers should consult [the OpenSAFELY Contracts
+reference](contracts-reference.md) to see:
+
+* what data is available in OpenSAFELY backends
+* how the data is structured
+* what that data represents
+
+!!! warning
+    OpenSAFELY Contracts apply only when accessing OpenSAFELY data using
```
sebbacon, line 39: How about something like "OpenSAFELY contracts are a concept introduced with Data Builder, the replacement for our legacy cohort-extractor system"

docs/contracts-intro.md
```diff
@@ -2,11 +2,159 @@
...
+    * verify which backends are compatible with their dataset definitions
+    * run data definitions on multiple different backends, without
+      having to write specific code for each individual backend
+* More consistent documentation that:
+    * describes the commonalities of the data offered by different backends
+    * notes where individual backends provide data that deviates from others
+
+## Common and Custom Contracts
+
+Data backends that are accessible via OpenSAFELY provide data tables. Each
```
sebbacon, line 79:
```suggestion
Data backends that are accessible via OpenSAFELY provide _data tables_ that can be referenced in _dataset definitions_. Each
```

docs/contracts-intro.md
sebbacon: I think our (my?) understanding of the classification of contracts has shifted since we first came up with this classification. I think the idea of "required" contracts is unecessary: we will have tooling for people to see which backends support the concepts *they're* interested in.

I think I now think of "common contracts" as ones that have been validated as expressing concepts that can be applied across two or more backends.  "custom contracts" are ones that have been implemented for only one backend.

It's possible even this distinction isn't particularly useful. Every time a backend wants to join opensafely, we will have a conversation with them about the contracts they will implement, and this conversation may cause us to bump the version of a contract already used by another backend.

On the other hand, we may want to make the fact we've never thought about the generalisability of a new contract really clear through categorisation, or namespacing (so if there's no 111Calls core contract, then the first time we implement it, we would choose "Graphnet/111Calls"; the second time, we might create a new "NHSE/111Calls" contract based on the Graphnet one, and perhaps somehow deprecate the Graphnet one.

I feel this is still not decided and I'd appreciate more feedback, more widely. For now perhaps we should weaken the classification here, and just explain that if a contract has not been validated as generalisable across multiple backends, it will be namespaced as such?

docs/contracts-intro.md
```diff
@@ -2,11 +2,159 @@
...
+    the details of specific contracts.
+
+Each OpenSAFELY Contract relates to a specific domain of health record
+data. The name of the Contract indicates the data domain covered by the
+Contract. Tables that satisfy these Contracts can be accessed in a
+dataset definition via [Data Builder's query language,
+ehrQL](ehrql-intro.md).
+
+An OpenSAFELY Contract provides the following information for each
+associated table:
```
sebbacon, line 140: I think we're missing an explanation of namespacing here. I explored it a bit in the most recent gdoc that lists some contracts and principles.

docs/contracts-intro.md
```diff
@@ -2,11 +2,159 @@
...
+* the **column's description**. The description summarises what the
+  column represents.
+* the **column's data type**. For instance, a column's data type might
+  be one of the following:
+    * integer type
+    * floating point type
+    * date type
+    * string type
+    * or some other type not in this list
+* the **additional constraints on the column's data values**, other than
```
sebbacon, line 153: We should mention that our plan is for these to be validated for researchers, so they can expect these conditions to always be true, but at the moment they're informational only.

docs/data-provider-integration.md
```diff
@@ -0,0 +1,86 @@
...
+!!! note
+
+    OpenSAFELY Contracts are still in an initial design and
+    implementation phase. We already have designs for additional
+    Contracts that will be implemented in future.
+
+If no existing Contract corresponds to a healthcare data domain that you
+wish to offer, [please email our technical team to
+discuss](mailto:tech@opensafely.org). This may involve the creation of a
+new [Custom Contract](contracts-intro.md#common-and-custom-contracts).
```
sebbacon, line 65: If we go with my earlier suggestion, then this would be more like "this may involve the creation of a contract namespaced only to your organisation or backend", I think

docs/data-provider-integration.md
```diff
@@ -0,0 +1,86 @@
...
+away. Researchers only need be concerned with specifying the data they
+want, not how to access it.
+
+If you are a new data provider, [please email our technical team to
+discuss Data Builder integration](mailto:tech@opensafely.org).
+
+!!! note
+
+    [Existing backend query implementations](https://github.com/opensafely-core/databuilder/tree/main/databuilder/backends)
+    are in Data Builder's GitHub repository.
```
sebbacon, line 84: Perhaps say there are two parts, writing a query engine for the database (currently MSSQL and Databricks) and writing code to map individual tables in the database to Contracts - and link to both those folders?

includes/glossary.md
```diff
@@ -39,3 +39,6 @@
 *[dataset definition]: specifies the details of data — for example, patients — that you want to include in your study. Used by Data Builder.
 *[ehrQL]: electronic health record query language. The language used by the Data Builder tool to extract cohorts
 *[OpenSAFELY Contracts]: provide a specification to make clear the data provided by different data backends, and to provide guidelines and tooling for data providers wishing to standardise their data for researchers
+*[backend]: one of several clinical databases that researchers might query via OpenSAFELY.
```
sebbacon, line 42: add "data provider" - an operator of a database (in UK data protection law, a _data processor_) that contains sensitive patient data; who also provides a secure infrastructure for running opensafely next to that data

Then "backend" is "an individual clinical database provided by a data provider" (I think)

StevenMaude: In the "Contracts domain modelling", we used "backend implementer", which I take to mean the same as "data provider".

Any opinions here on the preferable term?

sebbacon: Tough one. I'm not really keen on "backend" but for reasons I can't quite articulate. I'd say either canvas more widely (if you have appetite) or make a call yourself!

sebbacon: (maybe because it makes me think of bottoms)

docs/contracts-intro.md
StevenMaude: My understanding based on the current Data Builder code is that we are still using generic contract names. In that sense, the documentation covers how the system is right now.

I'm adding a note that mentions planned impending changes.

includes/glossary.md
StevenMaude: I'll go back to using "data provider" which was what I originally had down. I changed this because I'd spotted while reading more about Contracts.

docs/contracts-intro.md
StevenMaude: > I think I now think of "common contracts" as ones that have been validated as expressing concepts that can be applied across two or more backends. "custom contracts" are ones that have been implemented for only one backend.

This is how I understood it from re-reading. It's my half-remembering, along with not re-reading properly the first time that caused the initial confusion here. 

***

I had some related thoughts that don't really go anywhere, but will note them.

What does including "Custom Contracts" give you over not having it?

* It marks to researchers clearly which contracts are only supported by a single backend.
  * Counterpoint: let's say, in a ridiculous example, that we have 100 backends. Some of the Contracts might be "Custom" and only supported by one backend. Others might be "Common", but only supported by two backends. Is the extra concept and distinction useful here?
* We're trying to encourage data providers to consider structuring their data, where possible, to make things easier for researchers. Having the distinction means they can still provide additional data, or their idiosyncratic forms of existing data. But we're still clearly saying: "this is a special case".
* Likewise, if we're encouraging re-use of existing Contracts reduces our burden in having to concoct ad-hoc contracts for every individual backend.

To me, the distinction feels more related to communication and advocacy, and how you want to position that. And less so a technical decision (but I'm happy to be corrected).

***

All the above said, I think you could do the same by the Contract's name or namespace, if required. That makes it clearer to the person using such a Contract, avoiding needing some kind of separate label. 

And, as we haven't fully decided, for now, maybe it's worth de-emphasising the distinction and avoid introducing a separate concept. Instead:

* warn that some Contracts may be particularly specific to one backend
* warn that such Contracts will be labelled as such, somehow (to be determined)

docs/contracts-intro.md
```diff
@@ -2,11 +2,245 @@
...
+
+OpenSAFELY Contracts are authored by the OpenSAFELY platform developers:
+the Contract designs are informed by discussion and feedback from [data
+providers](data-provider-integration.md) and researchers.
+
+### How do OpenSAFELY Contracts fit into OpenSAFELY?
+
+Researchers write [dataset definitions](dataset-definition.md) to access
+data tables from OpenSAFELY backends. This is a simple example of a
+dataset definition that accesses a data table:
```
sebbacon, line 37: I think we should explain which _bit_ accesses the data table - i.e. in this case, it's `patients`.

Possibly we should just show these two lines

```py
from databuilder.tables import patients
year_of_birth = patients.date_of_birth.year
```

And explain that it accesses the `date_of_birth` field of the `patients` table?

docs/contracts-intro.md
```diff
@@ -2,11 +2,245 @@
...
+Dataset definitions are run using OpenSAFELY Data Builder. When running
+the dataset definition against live backends on the OpenSAFELY platform,
+the data requested will then be extracted.
+
+!!! note
+    The dataset definition is provided as an example only here.
+    You do not need to understand how to write a dataset definition
+    to follow the rest of this page.
+
+**Data tables in OpenSAFELY backends must relate to an OpenSAFELY
```
sebbacon, line 50:
```suggestion
**Data tables in OpenSAFELY backends are always related to an OpenSAFELY
```

docs/contracts-intro.md
```diff
@@ -2,11 +2,245 @@
...
+    * describes the commonalities of the data offered by different backends
+    * notes where individual backends provide data that deviates from others
+
+## How Contracts work
+
+Data backends that are accessible via OpenSAFELY provide *data tables*
+that can be referenced in *dataset definitions*. Each data table
+accessible via OpenSAFELY satisfies an OpenSAFELY Contract.
+
+Each data backend in OpenSAFELY can choose which of those Contracts to
```
sebbacon, line 94:
```suggestion
Data providers can make data available to OpenSAFELY  by choosing the Contracts they will satisfy.
```

docs/contracts-intro.md
```diff
@@ -2,11 +2,245 @@
...
+satisfy.
+
+!!! note
+    Refer to the [available OpenSAFELY data backends](data-backends.md)
+    to see which backends implement which Contracts.
+
+All Contracts share the:
```
sebbacon, line 101: I think "All Contracts share the same:" works better?

sebbacon (approved):
There's a bit of repetition I think, and the ordering could be tweaked, as you suggest, but the bones are good! :shipit: !

benbc (approved):
I'm happy once Seb's latest comments have been addressed. Thanks for taking so much effort over this, Steve.

StevenMaude:
I think this is now finally fine to merge as is. Thanks all for helpful feedback :partying_face:

Issue #493, opened 2021-11-23
Upgrade this repository to newer Python


StevenMaude:
This runs with ~Python 3.7~ Python 3.8, which is only getting security fixes now.

It also may have impact if we need to import the new cohort-extractor, which is only tested with Python 3.9.

Update: this is also a bit of a clunky usability issue, because you need:
* Python 3.7 to build or serve the content in this repository locally
* Python 3.10 for generating some content related to Data Builder

ghickman:
IIRC this is because cohortextractor (v1) has a requirement on 3.7.

@evansd or @bloodearnest might remember where the 3.7 requirement came from?

evansd:
cohortextractor-v1 actually runs against [3.8](https://github.com/opensafely-core/cohort-extractor/blob/1be28f4c9c8e05f5c08374f060db34ed2aa0bf9d/Dockerfile#L6). I don't think there's any reason not to upgrade this repo.

In terms of version compatibility the only thing which runs on a platform we don't control is [opensafely-cli](https://github.com/opensafely-core/opensafely-cli) and (purely because the CLI tool vendors it) [job-runner](https://github.com/opensafely-core/job-runner). Historically we've said that these need to be 3.7 compatible because, at the time, that's all some researchers had easy access to. It's probably past time to revist that policy. But in any case, we test on more recent versions too so there's no reason to keep the docs repo on an old version.

StevenMaude:
Ah, what a lovely, naive thought I had :laughing:

After trial, error and then searching around, it seems that the latest Python in Cloudflare Pages is Python 3.7 (which is turn is based on a Netlify repository).

Maybe Docker's an undocumented option for the build, though I can't find *anything* on the internet, referring to use of Pages with Docker.

StevenMaude:
Thanks to @tomodwyer, we switched the documentation site to use Render instead of Cloudflare Pages temporarily. This is until Cloudflare Pages add support for newer Python versions.

That change means we can move to at least Python 3.8 for the documentation site, so I've opened #651 which moves us to Python 3.8.

Ways to address this issue more comprehensively:

* ensure cohort-extractor, Data Builder and this repository all work with the same Python version
* remove the dependency of this repository on installing cohort-extractor and Data Builder, so that we only need install the MkDocs-related tooling

PR #605, opened 2022-02-19
Document how to Update Data Builder Docs

ghickman:
This adds a section on how to update parts of documentation which come from Data Builder.  It assumes the pipeline itself does not need updating, and that we'll document that elsewhere.  My intention was to mirror the old cohortextractor flow of "I have changed something in cohortextractor, how do I update the docs page?"

@StevenMaude – re that elsehwere, do you think that should be another section on this page or live in the relevant projects?

### DIFF
--- a/docs/updating-the-docs.md
+++ b/docs/updating-the-docs.md
@@ -29,8 +29,6 @@ the [documentation repository's
 README](https://github.com/opensafely/documentation#building-locally-and-testing)
 that details the use of `pip-compile` for this.
 
----8<-- 'includes/glossary.md'
-
 ## Making changes to the dataset definition examples
 You will need to check out the [documentation repo](https://github.com/opensafely/documentation) so you can run the example generating toolchain.
 
@@ -48,3 +46,28 @@ Examples are included in the markdown files using the [pymdown snippet notation]
 Make sure you commit both the changes to the `examples` and `databuilder/snippets` directories.
 
 If you are updating multiple examples it might be easier to the extraction step happen each time you change the python modules, you can do this with `just databuilder/watch-examples`.
+
+
+## Updating Data Builder backend and contract documentation
+
+!!! note "You will need both Python 3.7 and 3.10 on your PATH to make these updates"
+
+!!! note "This workflow is not supported, and has not been tested on Windows."
+
+Some of the Data Builder information in this OpenSAFELY documentation is generated from the [Data Builder source code](https://github.com/opensafely-core/databuilder).
+
+There are three components used to achieve this:
+
+* [Data Builder repository](https://github.com/opensafely-core/databuilder/)
+* [A mkdocs plugin](https://github.com/opensafely-core/mkdocs-opensafely-databuilder/)
+* [This documentation site](https://github.com/opensafely/documentation/)
+
+To update the [backend](data-backends.md) or [contracts](contracts-reference.md) pages:
+
+1. Make your changes in the [Data Builder repo](https://github.com/opensafely-core/databuilder/).  A new version of databuilder will be released once your changes are merged to `main`.
+1. In the [documentation repo](https://github.com/opensafely/documentation/) run `just databuilder/upgrade-databuilder` to update the version of databuilder installed in the subdirectory's virtualenv.  A current limitation is that a new Data Builder release is needed, before rebuilding the generated content.  Data Builder releases are only made for certain pull requests.
+1. Run `just databuilder/generate-docs` to update the `public_docs.json` file.
+1. Create a branch in the documentation repo and commit the `public_docs.json` file to it.
+1. Submit a PR for your changes.  Cloudflare will build a deployment of your branch if you don't want to test locally.
+
+---8<-- 'includes/glossary.md'


docs/updating-the-docs.md
```diff
@@ -48,3 +46,26 @@ Examples are included in the markdown files using the [pymdown snippet notation]
 Make sure you commit both the changes to the `examples` and `databuilder/snippets` directories.
 
 If you are updating multiple examples it might be easier to the extraction step happen each time you change the python modules, you can do this with `just databuilder/watch-examples`.
+
+
+## Updating Data Builder Backend and Contract documentation
```
iaindillingham, line 51: Nitpick: Titles in this section seem not to capitalize the first letter of each word. _Data Builder_ is an exception, because it's a proper noun (cf. common noun). By this reasoning, I think that _Backend_ and _Contract_ should be _backend_ and _contract_.

docs/updating-the-docs.md
```diff
@@ -48,3 +46,26 @@ Examples are included in the markdown files using the [pymdown snippet notation]
...
+
+Some of the Data Builder documentation is generated from parts of databuilder's codebase.
+
+There are three components used to achieve this:
+
+* [Data Builder](https://github.com/opensafely-core/databuilder/)
+* [A mkdocs plugin](https://github.com/opensafely-core/mkdocs-opensafely-databuilder/)
+* [This documentation site](https://github.com/opensafely/documentation/)
+
+To update the [Backend](data-backends.md) or [Contracts](contracts-reference.md) pages:
```
iaindillingham, line 63: Same nitpick as before 🙂

docs/updating-the-docs.md
```diff
@@ -48,3 +46,26 @@ Examples are included in the markdown files using the [pymdown snippet notation]
...
+
+## Updating Data Builder Backend and Contract documentation
+
+!!! note "You will need both Python 3.7 and 3.10 on your PATH to make these updates"
+
+Some of the Data Builder documentation is generated from parts of databuilder's codebase.
+
+There are three components used to achieve this:
+
+* [Data Builder](https://github.com/opensafely-core/databuilder/)
```
iaindillingham, line 59: We refer to this as "the Data Builder repo" below. Could we do the same here? It indicates what "Data Builder" means in this context; that is, the repository.

docs/updating-the-docs.md
```diff
@@ -48,3 +46,26 @@ Examples are included in the markdown files using the [pymdown snippet notation]
...
+* [A mkdocs plugin](https://github.com/opensafely-core/mkdocs-opensafely-databuilder/)
+* [This documentation site](https://github.com/opensafely/documentation/)
+
+To update the [Backend](data-backends.md) or [Contracts](contracts-reference.md) pages:
+
+1. Make your changes in the [Data Builder repo](https://github.com/opensafely-core/databuilder/).  A new version of databuilder will be released once your changes are merged to `main`.
+1. In the [documentation repo](https://github.com/opensafely/documentation/) run `just databuilder/upgrade-databuilder` to update the version of databuilder installed in the subdirectory's virtualenv.
```
iaindillingham, line 66: I think we should emphasize that this is done after a new version of databuilder is released. (Do all merges to `main` prompt a new release?)

iaindillingham (approved):
Nitpicks ⚠️ I'll approve, but @StevenMaude should also 👀.

docs/updating-the-docs.md
ghickman: I've not mentioned the repo here because it's not the repo per se, but the tool itself that is the component.  I linked to the repos for each component because that's where the one can make changes.  I don't feel strongly about it though, do you think it's worth changing still?

docs/updating-the-docs.md
ghickman: > Do all merges to main prompt a new release?

Yep!  The `build_and_deploy` workflow tags the repo based on the semantic commit used, and builds a new docker image.  Technically there's a race condition where you could hit merge, and then run `just databuilder/upgrade-databuilder` before the tagging flow finishes, but I suspect it's very unlikely to happen and I'm not sure it would be a valuable addition to the docs?

docs/updating-the-docs.md
iaindillingham: That's reasonable. At some point, it would be useful to note down what we're referring to with Data Builder, databuilder, etc. I don't feel strongly about it, though 🙂

ghickman: The [canonical name is `Data Builder`](https://app.shortcut.com/ebm-datalab/epic/372/rename-to-data-builder?vc_group_by=day&ct_workflow=all&cf_workflow=500000005), but I suspect many of us are referring to `databuilder` accidentally or to mean the repo without thinking about it!

docs/updating-the-docs.md
iaindillingham: I think I've misunderstood. The `main` workflow's `tag-new-version` step doesn't always create a new release (only for `fix`, `perf`, `feat`, and `BREAKING CHANGE`). If I merged a `docs` change to `main`, such as changing the docstring for `PatientDemographics`, then a new release wouldn't be created, so this step would be a noop, wouldn't it?

ghickman: TIL `docs` doesn't create a new release… arg.

I'll amend to make this part conditional, thanks for catching this!

iaindillingham: Thank you 🙂

iaindillingham: `github-tag-action` never fails to be confusing.

ghickman: Oh wait, we actually need a new version of databuilder to be built so we can use the tooling here because we only install a versioned build of it 😩

The alternative is to run the command with a local copy of the databuilder and then copy `public_docs.json` over, but I think that's going to make the documentation here very complicated.  However, the alternative feels like a foot gun waiting to happen 🤔

docs/updating-the-docs.md
```diff
@@ -48,3 +46,26 @@ Examples are included in the markdown files using the [pymdown snippet notation]
 Make sure you commit both the changes to the `examples` and `databuilder/snippets` directories.
 
 If you are updating multiple examples it might be easier to the extraction step happen each time you change the python modules, you can do this with `just databuilder/watch-examples`.
+
+
+## Updating Data Builder backend and contract documentation
+
+!!! note "You will need both Python 3.7 and 3.10 on your PATH to make these updates"
```
StevenMaude, line 53: suggestion:

* Maybe clarify briefly why you need both (linking to #493 might suffice)
* Maybe worth warning that this isn't supported for Windows as yet, if that's the case still. [^1]

[^1]: We could try Gitpod, but not sure how this would work out.

docs/updating-the-docs.md
```diff
@@ -48,3 +46,26 @@ Examples are included in the markdown files using the [pymdown snippet notation]
...
+
+Some of the Data Builder documentation is generated from parts of databuilder's codebase.
```
StevenMaude, line 55: thought: I suspect there's more effort on the reader's part when trying to work out whether "Data Builder" and "databuilder" are the same or different here. My opinion rather than fact, though.

I think I'd always favour "Data Builder" unless explicitly referring to actually code use — that is, `import databuilder`. For the case of the repository, I'd probably write "Data Builder repository" and link to the repository. Anyone reading, feel free to suggest counter-examples.

(We may have inconsistencies elsewhere and we can tidy these up, before we make the documentation more public.)

docs/updating-the-docs.md
StevenMaude: thought: I suspect there's more effort on the reader's part when trying to work out whether "Data Builder" and "databuilder" are the same or different here. My opinion rather than fact, though.

I think I'd always favour "Data Builder" unless explicitly referring to actually code use — that is, `import databuilder`. For the case of the repository, I'd probably write "Data Builder repository" and link to the repository. Anyone reading, feel free to suggest counter-examples.

(We may have inconsistencies elsewhere and we can tidy these up, before we make the documentation more public.)

docs/updating-the-docs.md
StevenMaude, line 55:
```suggestion
Some of the Data Builder information in this OpenSAFELY documentation is generated from the [Data Builder source code](https://github.com/opensafely-core/databuilder).
```

docs/updating-the-docs.md
StevenMaude: comment: I think this still need addressing :arrow_up:

StevenMaude: suggestion: The quick fix might just be to say:
* a current limitation is that you need to wait for a new release before rebuilding the generated content
* note that releases are only made for certain pull requests

StevenMaude:
summary: This looks mostly fine as is. I think it'll be subject to future refinement, in any case.

docs/updating-the-docs.md
StevenMaude: The Justfile seems bash-heavy. My assumption is that trying to run the same commands in Windows will fail.

StevenMaude:
> It assumes the pipeline itself does not need updating, and that we'll document that elsewhere.
> …
> re that elsehwere, do you think that should be another section on this page or live in the relevant projects?

Sorry @ghickman: I entirely skimmed this question on first look here :face_with_spiral_eyes:

To answer, my instinct is that such "pipeline" information could go further down this same page, if it might encompass *multiple different projects* where the outputs are in this repository. Perhaps with a warning that the guidance is more for software maintainers than researchers.

It's perhaps easier to see when there's a concrete example of what to add. In any case, I think it's fine not to bother considering that for this PR.

ghickman:
@StevenMaude – updated per feedback, could you take another look over it?

docs/updating-the-docs.md
```diff
@@ -48,3 +46,28 @@ Examples are included in the markdown files using the [pymdown snippet notation]
 Make sure you commit both the changes to the `examples` and `databuilder/snippets` directories.
 
 If you are updating multiple examples it might be easier to the extraction step happen each time you change the python modules, you can do this with `just databuilder/watch-examples`.
+
+
+## Updating Data Builder backend and contract documentation
+
+!!! note "You will need both Python 3.7 and 3.10 on your PATH to make these updates"
+
+!!! note "This workflow is not supported on Windows, it might work but we have not been able to test it"
```
StevenMaude, line 55:
```suggestion
!!! note "This workflow is not supported, and has not been tested, on Windows."
```

StevenMaude:
@ghickman: I added one more small suggestion to consider. Otherwise, this looks good to merge.

StevenMaude (approved):

PR #615, opened 2022-02-28
Add a minimal ehrQL example to the tutorial

iaindillingham:
I'll leave some comments that capture my thought process.

### DIFF
--- a/databuilder/snippets/ehrql.py
+++ b/databuilder/snippets/ehrql.py
@@ -1,0 +1,41 @@
+# :snippet minimal-ehrql-import-patients
+from databuilder.tables import patients
+
+year_of_birth = patients.date_of_birth.year
+# :endsnippet
+
+# :snippet minimal-ehrql-import-dataset
+from databuilder.query_language import Dataset
+
+dataset = Dataset()
+# :endsnippet
+
+# :snippet minimal-ehrql-set-population
+dataset.set_population(year_of_birth >= 2000)
+# :endsnippet
+
+# :snippet minimal-ehrql-extract-year-of-birth
+dataset.year_of_birth = year_of_birth
+# :endsnippet
+
+# :snippet minimal-ehrql-extract-birth-year
+dataset.birth_year = year_of_birth
+# :endsnippet
+
+# :snippet minimal-ehrql-register
+from databuilder.definition import register
+
+register(dataset)
+# :endsnippet
+
+# :snippet minimal-ehrql
+from databuilder.definition import register
+from databuilder.query_language import Dataset
+from databuilder.tables import patients
+
+year_of_birth = patients.date_of_birth.year
+dataset = Dataset()
+dataset.set_population(year_of_birth >= 2000)
+dataset.year_of_birth = year_of_birth
+register(dataset)
+# :endsnippet
--- a/docs/ehrql-tutorial.md
+++ b/docs/ehrql-tutorial.md
@@ -2,7 +2,50 @@
 
 ---8<-- 'includes/data-builder-danger-header.md'
 
-This page will contain an [ehrQL](ehrql-intro.md) tutorial. It will
-initially describe how a relatively minimal ehrQL example works.
+In this tutorial, we will walk through a minimal ehrQL example.
+The example will create a dataset of all patients who were born in or before 2000.
+We will walk through the example step-by-step;
+if you want to see the example in full, then you can [skip to the end](#wrapping-up).
+
+First, we import the patients table and create a variable that references each patient's year of birth:
+
+---8<-- 'examples/src-minimal-ehrql-import-patients.md'
+
+Next, we import and create a dataset
+
+---8<-- 'examples/src-minimal-ehrql-import-dataset.md'
+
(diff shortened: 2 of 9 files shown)


databuilder/snippets/ehrql.py
```diff
@@ -0,0 +1,41 @@
...
+dataset.birth_year = year_of_birth
+# :endsnippet
+
+# :snippet minimal-ehrql-register
+from databuilder.definition import register
+
+register(dataset)
+# :endsnippet
+
+# :snippet minimal-ehrql
```
iaindillingham, line 31: Maybe this should be above the other snippets, as they are derived from it. Maybe the file should also mirror `docs/ehrql-tutorial.md` and be called `databuilder/snippets/ehrql-tutorial.py`?

docs/ehrql-tutorial.md
```diff
@@ -2,7 +2,50 @@
 
 ---8<-- 'includes/data-builder-danger-header.md'
 
-This page will contain an [ehrQL](ehrql-intro.md) tutorial. It will
-initially describe how a relatively minimal ehrQL example works.
+In this tutorial, we will walk through a minimal ehrQL example.
+The example will create a dataset of all patients who were born in or before 2000.
+We will walk through the example step-by-step;
+if you want to see the example in full, then you can [skip to the end](#wrapping-up).
```
iaindillingham, line 8: I've used the first-person plural throughout the tutorial ("We..."). I've used the future form in the introduction ("We will walk through") but the present simple in the body ("We import").

docs/ehrql-tutorial.md
```diff
@@ -2,7 +2,50 @@
...
+
+First, we import the patients table and create a variable that references each patient's year of birth:
```
iaindillingham, line 10: I've not used variable, method, function etc. names in the body, as they are more brittle than the associated concepts. That is, `patients` is more brittle than "the patients table".

docs/ehrql-tutorial.md
```diff
@@ -2,7 +2,50 @@
...
+
+Next, we import and create a dataset
+
+---8<-- 'examples/src-minimal-ehrql-import-dataset.md'
+
+and set the population, which is all patients who were born in or before 2000:
+
+---8<-- 'examples/src-minimal-ehrql-set-population.md'
+
+At this point, we have created a dataset and set the population, but we haven't extracted any data.
```
iaindillingham, line 22: Technically, we have: `patient_id`. Sometimes, I think it's worth trading accuracy for clarity 🙂

docs/ehrql-tutorial.md
sebbacon: Important to get @StevenMaude's view on our "voice policy" here, though post-merge would be fine...

databuilder/snippets/ehrql.py
```diff
@@ -0,0 +1,45 @@
...
+
+# :snippet minimal-ehrql-import-dataset
+from databuilder.query_language import Dataset
+
+
+dataset = Dataset()
+# :endsnippet
+
+# :snippet minimal-ehrql-set-population
+dataset.set_population(year_of_birth >= 2000)
```
ghickman, line 16:
```suggestion
dataset.set_population(year_of_birth <= 2000)
```

We want births in and _before_ 2000, right?

docs/ehrql-tutorial.md
```diff
@@ -2,7 +2,50 @@
...
+Let's change that.
+As well as using each patient's year of birth to set the population, we also extract it:
```
ghickman, line 24: Are we actually extracting it here?  Given a dataset definition is more like a blueprint which a backend exercises to produce a dataset of patients, should we talk about extractions as something that happen in the future, and instead describe setting variables of on the dataset as an instruction to databuilder/the backend to get that data?

ghickman (commented):
Love seeing the examples working in this!

One comment on tone/wording/etc around explaining extraction, something to think on/discuss rather than a strong feeling!

databuilder/snippets/ehrql.py
iaindillingham: Actually, it's the other way around 🤦🏻‍♂️ The logic is correct (it's the same study as we use elsewhere), but the description in `docs/ehrql-tutorial.md` is not. Fixed in 06f684f

docs/ehrql-tutorial.md
iaindillingham: Good point! We're not, so I've changed "extract" to "request", except where we discuss tables and columns in the note. I chose "request" because I wanted a one word, regular verb to avoid lengthy statements, such as "we haven't instructed the backend which data we want it to write to the data file". Fixed in 9d67fcd

ghickman: > Dearest Research Backend, we would like to politely request…

iaindillingham:
If you're happy with the changes, would you mind ✅ @ghickman?

CarolineMorton (approved):
Great to see this moving on. Looks clear to me. I think we need to think about our voice as Seb pointed out but these small edits can occur after this merge and shouldn't block it.

PR #623, opened 2022-03-02
Add link to repo that joins ethnicity data in R

LindaNab:
I've only added the link to my repo now. Lmk if you think script should be added too. I did not add it because my script depends on input about the dates, which makes it a bit less standalone than the Python one. I could make one that works without inputting the dates tho.


LFISHER7 (approved):
Thanks for adding this, I think a link is fine for now!

PR #625, opened 2022-03-02
Fix toolchains' access to config

iaindillingham:
config is used by three toolchains: Cloudflare Pages (`.venv`), Data
Builder (`databuilder/.venv`), and pre-commit (the tool). To allow each
toolchain to access config, we ensure that flake8's per-file ignores are
relative to the project root, and we run the check and fix recipes in
databuilder/justfile from the project root.

You might be thinking "Cloudflare Pages doesn't need config; so why not
move config to Data Builder?" Well, then we'd break pre-commit (the
tool). You then might be thinking "Why not move pre-commit (the tool) to
Data Builder? Well, then everyone who cloned the repo would have to
setup both Cloudflare Pages and Data Builder to commit, even if the
commit was unrelated to Data Builder (see #619). As well as being
unnecessary, setting up both toolchains is a pain, because
`.python-version`, which is used by Dependabot, prevents pyenv from
putting 3.7 and 3.10 on the PATH.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -5,5 +5,5 @@ ignore = \
     W503   # line break before binary operator (black disagrees)
 max-line-length = 88
 per-file-ignores =
-    update-version.py:INP001
-    snippets/*:E402
+    databuilder/update-version.py:INP001
+    databuilder/snippets/*:E402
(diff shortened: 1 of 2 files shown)


bloodearnest (approved):
This is a hard problem, and this is about as elegant a solution to it as possible!

Issue #626, opened 2022-03-03
Improve the "Extracting ethnicity..." programming trick


iaindillingham:
My instinct says the `join_ethnicity` function in the [Extracting ethnicity across multiple time points][1] section is memory-intensive. I'd like to test it.

This programming trick also suggests overwriting input files. There are pros and cons (discussed in [this Slack thread][2]). I'd like to document them, too.

[1]: https://docs.opensafely.org/study-def-tricks/#extracting-ethnicity-across-multiple-time-points
[2]: https://ebmdatalab.slack.com/archives/C31D62X5X/p1646246157839709

PR #628, opened 2022-03-04
Rewrite "Extracting ethnicity..." section

iaindillingham:
I've attempted to summarize the one cohort/multiple cohorts strategy; and to update the scripted action to 1) make it more performant and 2) write out new files rather than overwrite existing files. (If you're interested in how we know that merging is more performant than mapping, then see [this gist](https://gist.github.com/iaindillingham/4903394b65dc3bad3b54e0eb1cde7ea5).)

The ability to overwrite existing files is, arguably, a bug in job-runner. If we're doing so to save disk space, then we should start by using compressed file formats (e.g. `.csv.gz`) first.


LFISHER7 (approved):
This looks much clearer. Thanks for the clear gist!

docs/study-def-tricks.md
```diff
@@ -321,63 +321,68 @@ study = StudyDefinition(
...
+    input_other = input_other.merge(input_ethnicity, how="left", on="patient_id")
+
+    # If the old path was "output/input_2021-01-01.feather", then the new
+    # path will be "output/input_2021-01-01_merged.feather".
+    new_path = path_to_output / f"{input_other.stem}_merged{input_other.suffix}"
+    input_other.to_feather(new_path)
 ```
 
-You can find an example of this in [this repo](https://github.com/opensafely/sro-template) (using Python) or in [this repo](https://github.com/opensafely/covid_mortality_over_time) (using R).
+You can find a Python example in [this repo](https://github.com/opensafely/sro-template) and an R example in [this repo](https://github.com/opensafely/sro-template).
```
StevenMaude, line 388: question: is the link to the R example correct here? I checked the linked repository but can't find any R code there (unless I'm missing something). Or is it supposed to be the [original link](https://github.com/opensafely/covid_mortality_over_time)?

StevenMaude: suggestion: Perhaps consider make the "this repo" link text more useful, especially as it's duplicated in the same sentence. For example:

```suggestion
### Research examples

Refer to existing research repositories using this extraction approach:

* [Python code](https://github.com/opensafely/sro-template)
* [R code](https://github.com/opensafely/covid_mortality_over_time).
```

The above change :arrow_up: assumes that the original link to the R code is correct, and the change to `sro-template` is a mistake.

StevenMaude:
@iaindillingham: did you have more to add here or do you want to merge? It looks pretty much done, from an onlooker's view.

While here, I also added a very small question.

StevenMaude:
@iaindillingham: And will merging this close #626, or is there more to do there?

PR #629, opened 2022-03-07
Introduce ehrQL

iaindillingham:
Leads the reader to the tutorial, examples, and reference. Written with reference to [the spec](https://docs.google.com/document/d/1jMEsjH2EbbGzrK-cAwCjSSRkQQ6JZu0YbckmPju2VHc/edit#heading=h.jifperpt4dpf). For more information, see [sc-735](https://app.shortcut.com/ebm-datalab/story/735/add-an-introduction-to-ehrql-to-the-documentation).

### DIFF
--- a/includes/glossary.md
+++ b/includes/glossary.md
@@ -35,7 +35,7 @@
 *[API]: Application Programming Interface. An agreed way for one system to communicate with another system.
 *[released outputs]: files from Level 4 uploaded to the Jobs site.
 *[draft published outputs]: a collection of released outputs prepared for publication to the general public.
-*[Data Builder]: a program that uses a data definition to create a datasets for analysis
-*[dataset definition]: specifies the details of data — for example, patients — that you want to include in your study. Used by Data Builder.
-*[ehrQL]: electronic health record query language. The language used by the Data Builder tool to extract cohorts
+*[Data Builder]: A program that uses dataset definitions to create datasets for analysis.
+*[dataset definition]: A file that specifies the criteria for selecting, and the characteristics of, the population. Used by Data Builder.
+*[ehrQL]: electronic health record query language. The language used by Data Builder to extract datasets for analysis.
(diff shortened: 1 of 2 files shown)


includes/glossary.md
```diff
@@ -35,7 +35,7 @@
 *[API]: Application Programming Interface. An agreed way for one system to communicate with another system.
 *[released outputs]: files from Level 4 uploaded to the Jobs site.
 *[draft published outputs]: a collection of released outputs prepared for publication to the general public.
-*[Data Builder]: a program that uses a data definition to create a datasets for analysis
-*[dataset definition]: specifies the details of data — for example, patients — that you want to include in your study. Used by Data Builder.
-*[ehrQL]: electronic health record query language. The language used by the Data Builder tool to extract cohorts
+*[Data Builder]: A program that uses data definitions to create datasets for analysis.
```
evansd, line 38: minor point: Should `data definitions` be `dataset definitions` for consistency below?

larger point: I wonder if the phrase "data/dataset definitions" needs to be here at all. There's a danger, I think, of having too many interrelated definitions and that getting confusing. Could we get away with something like?
```
*[Data Builder]: A program which fetches the data needed for an analysis
```

includes/glossary.md
```diff
@@ -35,7 +35,7 @@
...
+*[dataset definition]: A file that specifies the criteria for selecting, and the characteristics of, the population. Used by Data Builder.
+*[ehrQL]: The electronic health record query language. The language used by Data Builder to extract datasets for analysis.
```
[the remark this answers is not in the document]
iaindillingham: I'm afraid it just slipped in; I'll remove it. I wrote the sentence...

> _ehrQL_ (pronunciation), or the electronic health record query language, is the language...

...and didn't think about the "the". I then copied it to the glossary.

includes/glossary.md
iaindillingham: Good spot! It should be "dataset definitions" for consistency. Your larger point is also worth reflecting on: possibly as part of [sc-712](https://app.shortcut.com/ebm-datalab/story/712/update-glossary).

PR #637, opened 2022-03-09
Use python-legacy with mkdocstrings

ghickman:
This adds the `python-legacy` extra to mkdocstrings [as per its docs](https://github.com/mkdocstrings/mkdocstrings/blob/5867162b2eebf4829d39b6fe74843b60ced05a30/docs/handlers/overview.md#about-the-python-handlers) and turns off UserWarnings from that module.

### DIFF
--- a/justfile
+++ b/justfile
@@ -9,6 +9,13 @@ export PIP := BIN + "/python -m pip"
 export COMPILE := BIN + "/pip-compile --allow-unsafe --generate-hashes"
 
 
+# ignore the UserWarning emitted by mkdocstrings about explicitly using its
+# python-legacy extra.  Since this configuration can't be detected by the
+# library we have to ignore the warning.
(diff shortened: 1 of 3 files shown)


justfile
```diff
@@ -9,6 +9,13 @@ export PIP := BIN + "/python -m pip"
 export COMPILE := BIN + "/pip-compile --allow-unsafe --generate-hashes"
 
 
+# ignore the UserWarning emitted by mkdocstrings about explicitly using it's
```
StevenMaude, line 12: nit:

```suggestion
# ignore the UserWarning emitted by mkdocstrings about explicitly using its
```

StevenMaude (approved):
:+1:

Should we open an issue for testing the new handler, and migrating from `python-legacy`?

ghickman:
I'm not sure it's worth a ticket, I suspect we'll know if python-legacy stops working either by the build breaking or a new warning being emitted.

PR #646, opened 2022-03-14
Document date arithmetic errors with ceiling dates

rebkwok:
Document potential errors that can occur when using date arithmetic without filtering out ceiling dates (i.e. null values represented as 9999-12-31.  See also https://github.com/opensafely-core/cohort-extractor/issues/751

Also moved the date arithmentic docs to their own section as the current layout may imply they apply specifically to index dates

### DIFF
--- a/docs/study-def-dates.md
+++ b/docs/study-def-dates.md
@@ -16,6 +16,7 @@ As well as specifying dates explicitly with e.g., `on_or_before="2019-12-31"`, y
 You should be aware that events can be recorded in clinical systems without a date:
 
 - These default to "1900-01-01" in OpenSAFELY but other dates around and before this date are also possible. 
+- In some systems, null values for dates may be recorded with the ceiling value "9999-12-31".
 - These often relate to demographic information; and sometimes events or diagnoses (e.g. Asthma), which may have been imported from other systems, originally recorded on paper, or reported by patients relating to past experiences but not knowing the precise date. 
 - In addition, occasionally "impossible" dates may be recorded by accident; for example far in the future, or before a patient was born.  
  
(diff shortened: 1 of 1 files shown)


docs/study-def-dates.md
```diff
@@ -59,7 +60,25 @@ The available options (hopefully self-explanatory) are:
...
 Note that NHS financial (reporting) year runs from 1st April to 31st March.
 
-Intervals of time can be added or subtracted from the index date (or from a function applied to the index date).
+Any `index_date` you've defined in your study definition can be overridden in your `project.yaml`, by providing an `--index-date-range` argument, like this:
+
+```yaml
+actions:
+
+  generate_study_population_1:
+    run: cohortextractor:latest generate_cohort --study-definition study_definition --index-date-range "2020-01-01"
```
iaindillingham, line 69: How about...

```suggestion
    run: >
      cohortextractor:latest generate_cohort
        --study-definition study_definition
        --index-date-range "2020-01-01"
```

I didn't know it was possible to pass a single date to `--index-date-range`!

docs/study-def-dates.md
```diff
@@ -69,25 +88,31 @@ For example:
 "index_date - 1 year"
 ```
 
-Note that if the index date is 29 February and you add or subtract some number of years which doesn't lead to a leap year, then an error will be thrown.
-An error will also be shown if adding or subtracting months leads to a month with no equivalent day e.g. adding 1 month to 31 January to produce 31 February.
+Note that if the index date (or other starting date) is 29 February and you add or subtract some number of years which doesn't lead to a leap year, then an error will be thrown.
 
-Any `index_date` you've defined in your study definition can be overridden in your `project.yaml`, by providing an `--index-date-range` argument, like this:
+An error will also be shown if adding or subtracting months leads to a month with no equivalent day e.g. adding 1 month to 31 January to produce 31 February.
```
iaindillingham, line 93: Pedantry, pedantry...

```suggestion
An error will also be thrown if adding or subtracting months leads to a month with no equivalent day e.g. adding 1 month to 31 January to produce 31 February.
```

iaindillingham (approved):
Couple of nitpicks. But ✅

rebkwok:
> Couple of nitpicks. But white_check_mark

Nitpicks are in the bit of docs that I just moved and didn't re-read :slightly_smiling_face: but happy to include

iaindillingham:
Gotcha, sorry, I should have noticed that. Thank you 🙂

PR #651, opened 2022-03-16
Update documentation build to use Python 3.8

StevenMaude:
This does not completely solve #493. It does improve the situation slightly by removing the need for Python 3.7.


StevenMaude:
Apparently, the environment variable on Render was already `3.8.12`. So nothing to do there, I think.

CarolineMorton (approved):

PR #664, opened 2022-03-28
Technical architecture diagrams

lucyb:
System context and container diagrams, showing the different components
of OpenSafely and how they fit into the wider context. These diagrams
are written in the [C4 model](https://c4model.com/) style. They are
intended to assist with communication inside and outside of the team.


mkdocs.yml
```diff
@@ -18,6 +18,7 @@ nav:
               - Registered deaths: dataset-onsdeaths.md
               - Covid-19 therapeutics: dataset-therapeutics.md
       - Access policies: access-policies.md
+      - Tecnical architecture: technical-architecture.md
```
StevenMaude, line 21: nit:

```suggestion
      - Technical architecture: technical-architecture.md
```

StevenMaude:
questions:
(these are in a collected comment as no way to comment on the PNGs directly)

## System context diagram

* Does Graphnet need adding here?
* Is the lack of labels on the arrows intentional?

StevenMaude:
suggestions:

## Container diagram

* There's no label on the arrow from "Member of the public" to "OpenSAFELY Reports" while every other arrow has a label — maybe it should be "Accesses"?
* The GitHub container has a placeholder "Description of external software system" — does this need updating?
* Perhaps the description in OpenSAFELY Reports should be "Bennett Institute" instead of "DataLab"?

StevenMaude:
nit: I personally prefer imperative style for commit messages/PR titles. For example,

> Add technical architecture diagrams

but that's my personal preference, and not sure we have a team policy here.

StevenMaude (approved):

StevenMaude:
nit:

sorry, just spotted one more! "OpenCodeLists" should be "OpenCodelists".

lucyb:
> Is the lack of labels on the arrows intentional?

It was partly intentional to keep the diagram simple, but partly that I ran out of diagram editing steam at the time. Would it aid understanding of the diagram or is it simple enough that it's not needed? Genuine question, as I've been staring at it for too long now

lucyb:
> nit: I personally prefer imperative style for commit messages/PR titles. For example,

That's a good point. I tend to follow cbeams advice (https://cbea.ms/git-commit/), or I do when I remember. Generally, it seems like the team follows the same sort of convention though.

lucyb:
> While I'm here, what did you use for the diagrams?

Good point about having editable diagrams. I've not found a diagrams-as-code solution that I really like yet, so have been using diagrams.net (aka draw.io). It integrates with google drive, so I can share them that way, but then they're likely to get lost. I've just shared them with you via google, so you can see what I've been working on, but that's also not a particularly good long term solution. 

For now, I think I'll just export the SVGs as well and put them in the source code alongside the pngs (or instead of, depending on how they preview). I would welcome any better ideas though!

PR #671, opened 2022-03-30
Update various GitHub actions to v3

StevenMaude:


ghickman:
@StevenMaude – do you want to set up dependabot to do these next time as part of this change? [Example of that config from job-server](https://github.com/opensafely-core/job-server/blob/main/.github/dependabot.yml#L28-L31)

ghickman (approved):

StevenMaude:
@ghickman: TIL; will do, thanks!

PR #680, opened 2022-04-06
Reorganise code reviews section

StevenMaude:
This PR is a proof-of-concept attempt to reorganise the code
review documentation, applying the
[Diátaxis framework](https://diataxis.fr/).

This PR:

* breaks out the mixture of content from explanation and how-to.
* does some small rewriting, editing and tidying of content
* favours semantic line breaks (this is not part of the framework)

It is not entirely clear whether the "how" here follows the same
structure as a more strict how-to guide. Some of the content here
is more advice and guidelines rather than a more definite process.


StevenMaude:
Feedback welcome on this particularly :ear:

I'm as interested in what people think about the overall approach. Does it look like a better structure of this content?

madwort:
Just had a read through, followed by a look at the existing version. I like it - for me it has that feeling of overly long content that's been usefully refactored into chunks, making it easier to digest. And the Render preview was crucial for reviewing this.

StevenMaude:
One consideration here would be whether the URLs should be better organised or not. For instance, `code-reviews/for-authors` and `code-reviews/for-reviewers` might be better than `code-reviews-for-authors` and `code-reviews-for-reviewers`.

We haven't generally given too much thought to URLs, I don't think, but if we can have a good structure from the outset for new pages, that would be good.

PR #692, opened 2022-04-22
add background to case control page

alexwalkerepi:
Some background to describe the situation in which you'd use it.

### DIFF
--- a/docs/case-control-studies.md
+++ b/docs/case-control-studies.md
@@ -1,3 +1,7 @@
+## Background
+
+Data extracts often use an "index date" to define variables such as age, or presence of comorbidities at the start of a study. This can either be a fixed date (e.g. "2020-02-01") or a date that varies between patients, also known as [dynamic dates](https://docs.opensafely.org/study-def-dates/#dynamic-dates) (e.g. the date that people were hospitalised). However some studies that use dynamic index dates for a "case" population and then have a matched "control" comparator group. Because the control group by definition doesn't have the event of interest, the convention is for them to inherit the index date from their matched "case". The process to extract data for such a matched case-control population in OpenSAFELY is described below.
+
 !!! warning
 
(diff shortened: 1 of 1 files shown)


docs/case-control-studies.md
```diff
@@ -1,3 +1,7 @@
+## Background
+
+Data extracts often use an "index date" to define variables such as age, or presence of comorbidities at the start of a study. This can either be a fixed date (e.g. "2020-02-01") or a date that varies between patients, also known as [dynamic dates](https://docs.opensafely.org/study-def-dates/#dynamic-dates) (e.g. the date that people were hospitalised). However some studies that use dynamic index dates for a "case" population and then have a matched "control" comparator group. Because the control group by definition doesn't have the event of interest, the convention is for them to inherit the index date from their matched "case". The process to extract data for such a matched case-control population in OpenSAFELY is described below.
```
iaindillingham, line 3: Aside from adding a comma after "However", could I suggest using [Semantic Line Breaks](https://sembr.org/)? The tech team discussed this awhile back and added it to the [team manual](https://bennettinstitute-team-manual.pages.dev/tech-team/git-and-github-etiquette/#documentation-style).

```suggestion
Data extracts often use an "index date" to define variables such as age, or presence of comorbidities at the start of a study.
This can either be a fixed date (e.g. "2020-02-01") or a date that varies between patients, also known as [dynamic dates](https://docs.opensafely.org/study-def-dates/#dynamic-dates) (e.g. the date that people were hospitalised).
However, some studies that use dynamic index dates for a "case" population and then have a matched "control" comparator group.
Because the control group by definition doesn't have the event of interest, the convention is for them to inherit the index date from their matched "case".
The process to extract data for such a matched case-control population in OpenSAFELY is described below.
```

iaindillingham (approved):

PR #697, opened 2022-04-29
Two very minor formatting fixes

remlapmot:
If helpful spotted these. 

If you want formatting in the nav from mkdocs.yml you just need to code the html. And/or if you want the formatting in the title of actions-cohortextractor just include a level 1 header in that md file.


ghickman (approved):

ghickman:
Merging since the snippets check is a false negative here and fixed in #702

Thanks for these @remlapmot!

PR #708, opened 2022-05-05
Add first pass of notes for system integrators

StevenMaude:
Partially addresses [this documentation issue](https://github.com/opensafely-core/backend-server/issues/61).

See explanation in 2be926146df649c000c7ea9cdafa657bc05d8509 for more.

### DIFF
--- a/docs/system-integration.md
+++ b/docs/system-integration.md
@@ -7,14 +7,14 @@
 
 ## Audience
 
-This page is aimed at:
+This page is aimed at people who want to run their own installation of OpenSAFELY
+as a proof-of-concept trial prior to integrating into the live system.
+
+These people are likely to have one of the following professional roles:
 
 * software developers
 * clinical data providers
 * system integrators
-
-who want to run their own installation of OpenSAFELY
-as a proof-of-concept trial prior to integrating into the live system.
 
 ### Assumed knowledge
 
--- a/docs/system-integration.md
+++ b/docs/system-integration.md
@@ -1,0 +1,82 @@
+!!! warning
+
+    These notes are a work-in-progress.
+    This page provides an overview as guidance only,
+    and not a series of definite instructions.
+    We plan to further expand on this information.
+
+## Audience
+
+This page is aimed at:
+
+* software developers
+* clinical data providers
+* system integrators
+
+who want to run their own installation of OpenSAFELY
+as a proof-of-concept trial prior to integrating into the live system.
+
+### Assumed knowledge
+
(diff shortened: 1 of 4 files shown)


docs/system-integration.md
```diff
@@ -0,0 +1,82 @@
...
+## Audience
+
+This page is aimed at:
+
+* software developers
+* clinical data providers
+* system integrators
+
+who want to run their own installation of OpenSAFELY
+as a proof-of-concept trial prior to integrating into the live system.
```
CarolineMorton, line 17: I would prefer not to split the sentence before and after the list. Perhaps:

```suggestion
This page is aimed at people who want to run their own installation of OpenSAFELY
as a proof-of-concept trial prior to integrating into the live system. This group is likely to be: 

* software developers
* clinical data providers
* system integrators


```

docs/system-integration.md
```diff
@@ -0,0 +1,82 @@
...
+1. Deploy a [*job runner*](https://github.com/opensafely-core/job-runner) within your secure environment.
+   A minimal configuration simply runs Docker containers
+   and stores any resulting container output on a local disk.
+
+2. Deploy a [*job server*](https://github.com/opensafely-core/job-server)
+   that the *job runner* polls for jobs that end users request to be run.
+
+     It is possible to use our existing [*job server*](https://jobs.opensafely.org);
+     [contact us](how-to-get-help.md#data-providers)
+     if you would like us to configure this for you.
```
CarolineMorton, line 45: Query: Is this likely to be supported within our current time and developer constraints?

CarolineMorton (approved):
This looks great. Thank you for taking this on.

docs/system-integration.md
StevenMaude: I'm not sure, but it's what [@sebbacon said recently](https://github.com/opensafely-core/backend-server/issues/61#issue-1223798579), so that's why it's retained here:

> Here's text I've sent people previously
> …
> You would configure it to poll our ["job server"](https://github.com/opensafely-core/job-server/) for work requested by end users (you can use the one we provision - you'd need to get in touch for us to configure at our end).

docs/system-integration.md
StevenMaude: I think this suggestion is better, but I'll just bump the extra sentence to a new line to try and preserve (roughly) semantic line breaks.

docs/system-integration.md
StevenMaude: So, I'll leave as is for now.

PR #713, opened 2022-05-09
Explain what IMD ranking order means

rebkwok:


rebkwok:
> Looks good.
> 
> Is it worth linking out to the reference material? File 5 [here](https://www.gov.uk/government/statistics/english-indices-of-deprivation-2019)

I considered it - at least linking to the page, if not specific files - but then it would need to be updated when the IMD is updated, since that is specifically the 2019 data.  It's not that regularly updates - every 4 years or so - but the old pages don't get retired (the 2015 page is still there), so we wouldn't necessarily notice and we could be linking to the wrong data.

PR #720, opened 2022-05-11
Add link checking :link:

### LINKED ISSUE #642: Automate link checks
Looking around at other organisations using a "docs as code" approach, a link checker is a fairly standard piece of validation.

There may be other software around, but [lychee](https://github.com/lycheeverse/lychee) looks like it could be a good contender. lychee's README also [has a list of other competing software](https://github.com/lycheeverse/lychee#features).

There may be occasionally private links that we have to skip over. These should be few in number.

StevenMaude:
Fixes #642.

As discussed in #642, checking the links in the documentation source proved useful in finding some broken and outdated links. So, let's try adding this as a regular check. It typically takes around 40-50 seconds in GitHub Actions to run. Most of this is setup time. The link check takes around 10 seconds right now.

This PR:

* Adds [lychee-action](https://github.com/lycheeverse/lychee-action) manually via the Actions tab and on a weekly schedule.
  * We can later review whether to check on every PR.
* Fixes up a non-HTTP URL to HTTPS.
* Excludes some inaccessible URLs from checks; see 1c4198e4e4994b938400c9cd44a91ed322a37511 for reasoning.

If it proves useful and reliable here, we could later consider:

* Applying to other sites with considerable number of links (for example, https://opensafely.org, https://bennett.ox.ac.uk)
* Pulling this workflow out into a composite action or reusable workflow to avoid duplicating the configuration across multiple repositories.

### DIFF
--- a/.github/workflows/check_links.yml
+++ b/.github/workflows/check_links.yml
@@ -19,5 +19,6 @@ jobs:
         uses: lycheeverse/lychee-action@v1.4.1
         with:
           args: "--exclude-mail --require-https --verbose --no-progress './**/*.md' './**/*.html'"
+          fail: true
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
--- a/.github/workflows/check_links.yml
+++ b/.github/workflows/check_links.yml
@@ -8,6 +8,9 @@ on:
   schedule:
     - cron: "12 10 * * 2"
 
+permissions:
+  contents: read
+
 jobs:
   links:
     runs-on: ubuntu-latest
--- a/.github/workflows/check_links.yml
+++ b/.github/workflows/check_links.yml
@@ -1,0 +1,23 @@
+---
+name: 'Check links'
+
+# yamllint disable-line rule:truthy
+on:
+  pull_request:
+  workflow_dispatch:
+  schedule:
+    - cron: "12 10 * * 2"
+
+jobs:
+  links:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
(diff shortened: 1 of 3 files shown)


.github/workflows/check_links.yml
```diff
@@ -0,0 +1,27 @@
+---
+name: 'Check links'
+
+# yamllint disable-line rule:truthy
+on:
+  pull_request:
+  workflow_dispatch:
+  schedule:
+    - cron: "12 10 * * 2"
```
iaindillingham, line 9: Does the `workflow_dispatch` trigger allow the workflow to be started from the GH UI?

The workflow takes 40-60s to run, which is about the same length of time it takes to build the documentation. Am I right in thinking that when this PR is merged, then building the documentation would take something like 80-120s? If so, then could we _not_ run it on each PR, accepting that a few links might be broken between scheduled runs?

iaindillingham (approved):

.github/workflows/check_links.yml
StevenMaude: Yes, `workflow_dispatch` is if you want to run manually for any reason.

The workflows here should all run concurrently, as far as I can tell, provided the organisation doesn't hit its limit of [concurrent jobs (60)](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits).

That all said: if we want to make it minimally irritating for contributors (potentially getting a failure on a PR via an existing broken link), let's try it just on a schedule first, and then add it on all PRs if we think it's useful.

StevenMaude: Updated in 4cf5922c69fa520929786a20d0d4b9313bf840ec.

PR #728, opened 2022-05-17
Fix quotes in scheduled workflow run

StevenMaude:
They must be
[single quotes](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule)
because this is YAML.


StevenMaude:
Noticed because this didn't trigger as it should today.

iaindillingham (approved):
yaml, yaml, yaml. Really? REALLY?

ghickman:
@StevenMaude – I'm fairly sure YAML doesn't treat quote type as significant, so I don't think that's the issue here.  I've got several scheduled workflows using double quotes and they're all triggering as expected, but I reckon it's still worth a try?

iaindillingham:
If it helps, I do everything yaml in VS Code and have it reformat everything.

ghickman:
I've also seen some folks tweeting about jobs not starting and I see that 4 of my scheduled jobs (all in one repo) are currently queued.

StevenMaude:
:man_facepalming:

StevenMaude:
check-run

ghickman:
It's always ~DNS~ timezones…

StevenMaude:
I've merged this because I think single quotes is actually the better thing to do here style-wise, though it doesn't fix anything.

Amended the commit message and PR message to reflect my thinking.

PR #754, opened 2022-05-26
Pages deployment

tomodwyer:
WIP 🚧 

Waiting on a fix for this: https://github.com/cloudflare/pages-action/issues/8

### DIFF
--- a/.github/workflows/pages-deployment.yaml
+++ b/.github/workflows/pages-deployment.yaml
@@ -11,11 +11,13 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Install dependencies
+      - name: Set up Python
         uses: actions/setup-python@v3
         with:
           python-version: "3.8.12"
           cache: "pip"
+
+      - name: Install dependencies
         run: pip install -r requirements.txt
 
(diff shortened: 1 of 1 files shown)


.github/workflows/pages-deployment.yaml
```diff
@@ -0,0 +1,33 @@
+on: [push]
```
StevenMaude, line 1: thought: If this happens on every commit, then I think that the existing `build.yml` is no longer required.

iaindillingham:
I have a minor gripe with Render, in that is has its own settings for building the documentation; these are independent of the "build-check" GitHub action.

I noticed that the documentation wasn't building, today. This was partly related to the Python handler. And partly to building in strict mode. And partly to not realizing that changing the settings in the "build-check" GitHub action wouldn't also change the settings for Render. #763 was the result. Exasperation in [this Slack thread](https://bennettoxford.slack.com/archives/C01SE1WB6AW/p1654096270518129).

.github/workflows/pages-deployment.yaml
tomodwyer: Good point, I will remove as part of this PR.

PR #763, opened 2022-06-01
Switch Python handler

iaindillingham:
Hopefully this will resolve an issue with pinned versions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,4 +26,4 @@ jobs:
           python3 -m pip install -r ./requirements.txt
 
       - name: Check site builds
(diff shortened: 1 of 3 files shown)


StevenMaude:
Also should resolve the problem with PR #757.

StevenMaude:
I've just skimmed the variable reference, which is the main thing we use mkdocstrings for.

The new version has added a `Type` column everywhere which is empty. I don't think this should block this PR, as the site currently doesn't deploy, but we should open issues for that, and for getting rid of the warnings we don't want that griffe gives us, so we can re-enable `--strict` in builds.

StevenMaude (approved):

PR #766, opened 2022-06-02
Update databuilder docs

inglesp:


iaindillingham:
Between cc7d662e25d7a21bc414524ce9d57fc4887dde76 and 314c32701f99a77485d8e873517251966c556e7e we switched from Render to Cloudflare Pages. CP is failing because it's building against Python 3.7 rather than 3.8. I thought this was why we moved from CP to Render. More [here](https://bennettoxford.slack.com/archives/C63UXGB8E/p1654514674763189).

PR #786, opened 2022-06-06
Add IMD quintiles trick

iaindillingham:
This commit complements #761.

### DIFF
--- a/docs/study-def-tricks.md
+++ b/docs/study-def-tricks.md
@@ -366,11 +366,11 @@ study = StudyDefinition(
     imdQ5 = patients.categorised_as(
         {
             "0": "DEFAULT",
-            "1": "imd >=0 AND imd < 32844*1/5",
-            "2": "imd >= 32844*1/5 AND imd < 32844*2/5",
-            "3": "imd >= 32844*2/5 AND imd < 32844*3/5",
-            "4": "imd >= 32844*3/5 AND imd < 32844*4/5",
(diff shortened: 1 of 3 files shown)


iaindillingham:
Here's the new trick: https://opensafely-documentation-pr-786.onrender.com/study-def-tricks/#grouping-imd-into-quintiles

docs/study-def-tricks.md
```diff
@@ -353,3 +353,35 @@ In addition:
...
+
+To group IMD into quintiles, then:
+
+```py
+study = StudyDefinition(
+    # ...
+    imdQ5 = patients.categorised_as(
+        {
+            "0": "DEFAULT",
+            "1": "imd >=0 AND imd < 32844*1/5",
```
rebkwok, line 369: Should we add a note that 32844 is the total LSOAs at the time of the IMD rankings currently in TPP (2019 data I think); it could change with the next update, and study defs would need to update this number

rebkwok (approved):
LGTM, just one question

docs/study-def-tricks.md
iaindillingham: 62c9814 replaces approximate values with actual values for both original and rounded rankings. I've also linked to [_English indices of deprivation 2019_](https://www.gov.uk/government/statistics/english-indices-of-deprivation-2019), which describes the number of LSOAs in England.

rebkwok (approved):

PR #789, opened 2022-06-08
Use more explicit names for IMD groups

wjchulme:
Typically users have coded IMD groups with integers, but there's no reason for the group names not use more descriptive strings. This approach also prevents grouped IMD being used as a purely numeric value downstream.


iaindillingham:
I've resolved the merge conflict and force-pushed @wjchulme. Thanks for clarifying!

iaindillingham (approved):

iaindillingham:
I'm going to merge, so I can update the links in the associated issue.

PR #792, opened 2022-06-08
Ignore `www.icnarc.org` links

StevenMaude:
And also refactor and correct the `.lycheeignore` slightly.

This makes it more consistent with the one for `opensafely.org`.

When intending to handle multiple related URLs:

* match the URL
* match any subpath of that URL
* match only the specific subdomain


StevenMaude:
https://github.com/opensafely/documentation/actions/runs/2461527665 :heavy_check_mark:

iaindillingham (approved):

PR #730, opened 2022-05-18
Add Linux instructions (fairly brief)

mdvthu:
Also removed some (? unintended) whitespace

PR #806, opened 2022-06-15
Fix whitespace in Getting Started guide

StevenMaude:
As suggested in #730 by @mdvthu.


StevenMaude:
I've broken this out as #730 could do with a little further work.

Issue #807, opened 2022-06-15
Dependabot version updates are failing in `databuilder/` directory


StevenMaude:
```

updater | ERROR <job_392265894> Error processing snex (Dependabot::SharedHelpers::HelperSubprocessFailed)
updater | ERROR <job_392265894> Using indexes:
updater | <job_392265894>   https://pypi.org/simple
updater | <job_392265894> 
updater | <job_392265894>                           ROUND 1                           
updater | <job_392265894> Traceback (most recent call last):
updater | <job_392265894>   File "/usr/local/.pyenv/versions/3.8.12/bin/pip-compile", line 8, in <module>
updater | <job_392265894>     sys.exit(cli())
updater | <job_392265894>   File "/usr/local/.pyenv/versions/3.8.12/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
updater | <job_392265894>     return self.main(*args, **kwargs)
updater | <job_392265894>   File "/usr/local/.pyenv/versions/3.8.12/lib/python3.8/site-packages/click/core.py", line 1055, in main
updater | <job_392265894>     rv = self.invoke(ctx)
updater | <job_392265894>   File "/usr/local/.pyenv/versions/3.8.12/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
updater | <job_392265894>     return ctx.invoke(self.callback, **ctx.params)
updater | <job_392265894>   File "/usr/local/.pyenv/versions/3.8.12/lib/python3.8/site-packages/click/core.py", line 760, in invoke
updater | <job_392265894>     return __callback(*args, **kwargs)
updater | <job_392265894>   File "/usr/local/.pyenv/versions/3.8.12/lib/python3.8/site-packages/click/decorators.py", line 26, in new_func
updater | <job_392265894>     return f(get_current_context(), *args, **kwargs)
updater | <job_392265894>   File "/usr/local/.pyenv/versions/3.8.12/lib/python3.8/site-packages/piptools/scripts/compile.py", line 466, in cli
updater | <job_392265894>     results = resolver.resolve(max_rounds=max_rounds)
updater | <job_392265894>   File "/usr/local/.pyenv/versions/3.8.12/lib/python3.8/site-packages/piptools/resolver.py", line 175, in resolve
updater | <job_392265894>     has_changed, best_matches = self._resolve_one_round()
[...]

PR #808, opened 2022-06-15
Add `.python-version` to `databuilder/` directory

StevenMaude:
It requires a different version.

The documentation around how Dependabot works is not great; maybe it
will pick this up as an override. It picks up that the requirements are
separate.


StevenMaude:
This is an attempt to deal with #807. If you're reading this and already know that this won't work, feel free to close.

StevenMaude:
Actually, this should be [Python 3.10](https://github.com/opensafely/documentation/blob/7e1595b123e50c6b29396148e23da25ac6ea2676/databuilder/justfile#L25).

ghickman (approved):
🤞

PR #810, opened 2022-06-16
Add some initial Linux installation notes

StevenMaude:
Prompted by @mdvthu in #730.

These notes are deliberately left a little to the reader to decide how
they want to proceed. This is for the reasons given in the text.

### DIFF
--- a/docs/install-linux.md
+++ b/docs/install-linux.md
@@ -1,0 +1,86 @@
+# Linux Install Guide
+
+This is a series of guidelines for installing the OpenSAFELY CLI on Linux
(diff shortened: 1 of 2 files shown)


StevenMaude:
I spotted a couple of things with the installation guides which could be improved:

1. They might be promoted to an upper level in the documentation hierarchy.
2. There is some overlap and duplication between installations described per operating system (Windows, macOS, Linux), and per piece of software (Git, Docker, OpenSAFELY CLI). We should tidy this in future.

For now, I think it's fine to leave as is, and resolve collectively as an entire section.

Issue #762, opened 2022-06-01
Add a date or commit ID somewhere to published HTML


StevenMaude:
@iaindillingham spotted that the docs weren't being published.

We only had an email notification of failure. Render doesn't seem to display a publish check here on GitHub. (I checked on Render and not sure there's anything there to configure.)

Cloudflare Pages does display status here, for when we switch back, #754, but it might still be useful to have an easy validation in-page that content is current. And useful for readers too, as a warning that content may be old.

StevenMaude:
A review date, [as the NHS suggest to have](https://service-manual.nhs.uk/design-system/patterns/reassure-users-that-a-page-is-up-to-date), might also be nice, but not sure if MkDocs for Material supports this.

StevenMaude:
We can probably pick this up again now as we've switched away from Render now in #921 and #923.

PR #817, opened 2022-06-22
Display page update from last Git revision

StevenMaude:
Relates to #762.

Using Material for MkDocs support for
`mkdocs-git-revision-date-plugin`.


StevenMaude:
I had to open this to test it.

Unfortunately, though the plugin itself works fine, Render appears to do a shallow clone of the repository. This breaks the collection of dates from the Markdown files. And there's no way to configure this checkout that I can see in the settings.

(Cloudflare Pages behaves similarly; at least it did in 2021.)

PR #820, opened 2022-06-23
feat: add warning re  limitations of consultations

Jongmassey:
re: https://github.com/opensafely-core/opensafely-cli/issues/120 I've added an extra "health warning" to this function. Whilst the limitations were already present in the documentation, this re-states them more strongly within a formatted warning box

### DIFF
--- a/docs/study-def-variables.md
+++ b/docs/study-def-variables.md
@@ -33,6 +33,8 @@ These variables are derived from data held in the patients' primary care records
 ::: cohortextractor.patients.with_gp_consultations
 &nbsp;
 
+---8<-- 'includes/gp-consultations-warning-header.md'
+
 ::: cohortextractor.patients.sex
 &nbsp;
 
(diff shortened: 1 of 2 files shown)


sebbacon:
Why render this as a snippet? (I can't see it being reused elsewhere).

Jongmassey:
I was following a pattern for [other](https://github.com/opensafely/documentation/blob/main/includes/imd-warning-header.md) [such](https://github.com/opensafely/documentation/blob/main/includes/isaric-warning-header.md) [warnings](https://github.com/opensafely/documentation/blob/main/includes/onscis-warning-header.md).

The alternative is to update the docstring in cohortextractor and let it percolate through from there.

Jongmassey:
That and the markdown for the warning box would have to be added to the docstring which feels a bit ugly.

Perhaps re-using the note in the [TPP SystmOne dataset](https://docs.opensafely.org/dataset-systmone/) page?

StevenMaude:
> That and the markdown for the warning box would have to be added to the docstring which feels a bit ugly.

I don't think this in itself would be too bad. Provided it indeed filters through to being rendered correctly in the docs (I can't remember off-hand).

*But*, that said, it looks like we currently just have warnings in this docs page, so I'd say this is fine to include in the page as is, since that's the current convention.
***

I think there is also a question of: if there are things here that warrant warnings on this page, would someone purely looking by reference to the cohort-extractor source know about them? How do researchers typically look up details of this API? Via this page, or the source?

This is all perhaps more a question for Data Builder though.

iaindillingham (approved):

PR #829, opened 2022-06-28
Add new page about federation

sebbacon:
Also sync precommit version with .python-version, so I can commit it


evansd (approved):
Looks great to me, for what that's worth!

docs/federation.md
```diff
@@ -0,0 +1,93 @@
...
+
+After execution, the **outputs were checked** for disclosivity by privacy specialists. When the results of the regression models were validated as safe to release, they were  **released** from the secure backend for direct access by the researcher. You can see the [TPP and EMIS outputs for the vaccination paper here](https://github.com/opensafely/Factors-associated-with-COVID-19-vaccination/tree/master/released_outputs) in their raw form.
```
StevenMaude, line 57: question:

> You can see the [TPP and EMIS outputs for the vaccination paper here](https://github.com/opensafely/Factors-associated-with-COVID-19-vaccination/tree/master/released_outputs) in their raw form.

Should this link be pinned to a commit ID, instead of just `master`? It avoids the link breaking if anyone does move the files around in future.

docs/federation.md
```diff
@@ -0,0 +1,93 @@
...
+
+Finally, the researcher combined the model outputs from both backends in a **meta-analysis**  with a [final script, which you can read here](https://github.com/opensafely/Factors-associated-with-COVID-19-vaccination/blob/master/analysis/R/Scripts/06_metaanalysis.R).  In this case, they combined model coefficients using inverse-variance-weighting.
```
StevenMaude, line 59: nit:

```suggestion
Finally, the researcher combined the model outputs from both backends in a **meta-analysis** with a [final script](https://github.com/opensafely/Factors-associated-with-COVID-19-vaccination/blob/master/analysis/R/Scripts/06_metaanalysis.R). In this case, they combined model coefficients using inverse-variance-weighting.
```

StevenMaude:
I've added a couple of small comments for polish :sparkles: and I don't think the `.pre-commit-config.yaml` should change here.

But no issues with the actual written content which looks good :+1:

StevenMaude (approved):

docs/federation.md
sebbacon: I can see it both ways, but on balance I think probably it's better for it to point to `master` in case there are important updates in the future

docs/federation.md
```diff
@@ -0,0 +1,93 @@
...
+
+OpenSAFELY was designed for federation from the start. It is installed as a data access layer in front of multiple existing patient databases; we call an installation a **backend**. For example, Primary Care data in England is provided by [two backends described here](dataset-intro.md), with the support of the two largest providers of primary care software in the country: TPP and EMIS.  In this way, researchers have safe access to the records of nearly 60 million patients in England.
+
+One problem OpenSAFELY solves is that backends may present very different environments and databases: for example, TPP provides their data in a _Microsoft SQL Server_ database on a hardware server, and EMIS provides it via their _EXA_ analytic platform in the cloud.  OpenSAFELY tools hide these differences from researchers, so that their code can be run **everywhere**. After code is run on a backend, its outcomes are carefully checked for safety within the same secure perimeter, and finally released so that they can be **recombined**:
+
+```mermaid
```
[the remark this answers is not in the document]
sebbacon: I've noticed an unpleasant FOUC type effect with mermaid; I wonder if hosting the javascript ourselves would solve this.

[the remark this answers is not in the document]
sebbacon: What do you think about the [privacy plugin](https://github.com/squidfunk/mkdocs-material/issues/3781#issuecomment-1082091621) as a solution? I'm quite keen to pay for mkdocs given how much use we've made of it, anyway

PR #832, opened 2022-06-29
Remove the need for installing cohort extractor

### LINKED ISSUE #493: Upgrade this repository to newer Python

### LINKED ISSUE #807: Dependabot version updates are failing in `databuilder/` directory

### LINKED ISSUE #830: Enable pre-commit checks for Python code not in the Data Builder directory
The current [.pre-commit-config.yaml](https://github.com/opensafely/documentation/blob/54d0f8cca83540a04ac5b913c948f2a439487d8e/.pre-commit-config.yaml) uses Python 3.10 and only checks the `databuilder/` directory.

It would be good to ensure that any Python code we add anywhere is checked.

The problem is that the main documentation repository still needs Python 3.8 (#493).

StevenMaude:
This may not be an ideal approach, but works.

The `opensafely-cohort-extractor` Python package dependency is replaced with a submodule dependency. Currently, we only need the docstrings from this code: we don't run any cohort-extractor related code.

This should:

* fix #493
* fix #807
* fix #830
* remove a pain point for people trying to install this repository locally and requiring two Python versions

It also removes a considerable number of dependencies, for building the documentation.

If we want to test cohort-extractor snippets in future, then we might need a solution for that. It may be possible to more tightly contain those checks to a snippets directory, and run in a separate virtualenv to that used to build the docs.

See the individual commit messages for explanation of some decision making.

This may not be an ideal approach, but works.

What this doesn't cover:

* Moving to Python 3.10 as Data Builder uses 3.9, and unsure if there are any problems there. The core problem is the two different versions; it would be easy enough to later to upgrade again.

See the individual commit messages for explanation of some decision making.

## Tasks

- [x] Document the use of submodules here, maybe with a brief rationale/link to this PR
- [x] Add a means to the `Justfile` to clone the submodule


.pre-commit-config.yaml
```diff
@@ -1,7 +1,5 @@
 default_language_version:
-  python: python3.10
-
-files: "^databuilder/"
+  python: python3.9
```
StevenMaude, line 2: question: Can we remove this specification entirely?

mkdocs.yml
```diff
@@ -7,7 +7,7 @@ nav:
       - Security: security-levels.md
       - Data sources:
           - data-sources/index.md
-          - Overview: data-sources/intro.md 
+          - Overview: data-sources/intro.md
```
StevenMaude, line 10: note: These whitespace changes are now picked up by pre-commit, so we have to fix them to placate it.

.pre-commit-config.yaml
ghickman: We need to set a python version, it can either be done per rule or with a default.  We've been moving to the default pattern everywhere of late as we've not found a need to change version per rule.

ghickman (commented):
Not sure how I feel about this!

I used submodules for years to manage m vim plugins and came to really dislike them.  Having dependabot handle the upgrades seems like it would remove many of the issues though.  Could we document how to manually bump the submodule too?

Is there any reason we need to stick to 3.9 in this change?  Bumping to 3.10 would mean we can drop all the `.python-version` files.

StevenMaude:
Using 3.10 is fine. The only reason I didn't was to initially get everything synced up, and that Data Builder itself is [only tested with 3.9](https://github.com/opensafely-core/databuilder/blob/d62660da28f88cfccc6112b1462b715256ed4e18/.github/workflows/main.yml#L30). 

Submodules can be sticky, _but_:

* I think it's a not entirely invalid use case here. All we want is a copy of the source to process the docstrings, not to actually run the code. The current mkdocstrings docs [actually advocate against installation into the virtualenv](https://mkdocstrings.github.io/python/usage/#installing-your-package-in-the-current-python-environment).
* The current situation has caused a number of frustrations, and it resolves them for now.

I'd also need to explain how to clone, update and probably add something to the `Justfile` to update the submodule.

ghickman:
@StevenMaude – agreed on both points and my hope is that the automation will remove the vast majority of submodule pain (and a little bit of me hopes that they got even a tiny bit better in the ~6y since I used them in anger!).

So I reckon go for it 👍

bloodearnest:
Seagulling in to suggest a potential completely alternate approach that avoids git submodules (which scare me).

I suggest we make the default databuilder package have no (or minimal) [requirements](https://github.com/opensafely-core/databuilder/blob/main/pyproject.toml#L14) by default, and we add `extras` in the package tooling to install the rest of stuff as needed e.g. ` pip install databuilder[full]`, which would bring in all the actual dependencies.

This is also a horrible hack. But for my money its better than submodules.

evansd:
@bloodearnest We were already planning to do the non-hacky version of that package split with databuilder anyway. I think this submodules approach is just about dealing with cohortextractor, and in that sense it seems like an improvement on the status quo.

StevenMaude:
Since Data Builder uses Python 3.9, and we do currently need to install it as a package, I've opted to stick with Python 3.9 for now, for simplicity, and to reduce the risk of any gotchas in future.

It would be much easier now to consider upgrading to 3.10 though.

StevenMaude:
Depending on when this actually gets merged, it could warrant a final rebase on `main` to ensure we don't downgrade any requirements.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "documentation"
-requires-python = ">=3.8<3.9"
+requires-python = ">=3.9<3.10"
```
ghickman, line 3: Does ~=3.9 work here? I've never actually tried pip-compatible version specifiers for this argument, but it might do?

IIRC I added this to try and get dependabot to pick up a version without using .python-version, so the `project` key can probably go anyway.

justfile
```diff
@@ -86,6 +86,13 @@ upgrade env package="": virtualenv
     test -z "{{ package }}" || opts="--upgrade-package {{ package }}"
     FORCE=true {{ just_executable() }} requirements-{{ env }} $opts
 
+# Fetch cohort-extractor submodule
+fetch-cohort-extractor:
+    git submodule update --init
+
+# Update cohort-extractor submodule
+update-cohort-extractor:
+    git submodule update --remote src/cohort-extractor
```
ghickman, line 95: 👌

INSTALL.md
```diff
@@ -8,47 +8,72 @@ in their docs includes lots of very nice customisations for formatting documenta
 
 ## Building locally
 
+- `git clone --recurse-submodules git@github.com:opensafely/documentation.git`
+  - `--recurse-submodules` ensures you have the necessary copy of the `cohort-extractor` source
 - Create a virtual environment
 - `pip install -r requirements.txt`
 - `mkdocs serve`
```
ghickman, line 15: Is this whole section superseded by `just run`? and if not, it probably should be to match all our other infra?

INSTALL.md
```diff
@@ -8,47 +8,72 @@ in their docs includes lots of very nice customisations for formatting documenta
...
+### Updating cohort-extractor manually
 
 Alternatively, you can pull in the latest version of the cohortextractor
 docstrings, for local development or to update the requirements entirely
 by hand:
 
-    pip-compile --allow-unsafe --generate-hashes -P opensafely-cohort-extractor
-    pip install -r requirements.txt
+    git submodule update --remote src/cohort-extractor
+    git add src/cohort-extractor
```
ghickman, line 49: Would it make sense to wrap this in a just command?

INSTALL.md
```diff
@@ -8,47 +8,72 @@ in their docs includes lots of very nice customisations for formatting documenta
...
 
-opensafely-cohort-extractor is a documentation dependency. We use the
-opensafely-cohort-extractor docstrings to generate some content here.
+[cohort-extractor](https://github.com/opensafely-core/cohort-extractor) is a documentation dependency.
+We use the cohort-extractor docstrings to generate some content here.
+
+**We currently do not install cohort-extractor into a virtualenv
+because we are only using the docstrings**
+cohort-extractor is currently a Git submodule instead,
+for installation simplicity (see #832).
```
ghickman, line 25: Nit: I don't love the intra-sentence line breaks here.  I know sembr suggests it but it bring me no joy to read! YMMV 🙂

INSTALL.md
```diff
@@ -8,47 +8,72 @@ in their docs includes lots of very nice customisations for formatting documenta
...
+
+We hope to make this easier to work with in future:
+
+#### Current state
+
+Rather than limit Data Builder to the versions supported by Cloudflare Pages, we've opted to have it generate a JSON file from Data Builder and commit that JSON to this repo.
+
+Run `just databuilder/generate-docs` to generate the JSON file using `databuilder` installed in the `databuilder` subdirectory of this repo.
+
+If the JSON file has changed you might need to update [the plugin which uses it to render pages](https://github.com/opensafely-core/mkdocs-opensafely-backend-contracts/).
```
ghickman, line 79: I'm a bit confused by these two sections!

In the first one we have a line starting `For now`, but it's a historical context, so maybe that should be in the second section?

Then there's the line ` We hope to make this easier to work with in future:` with a new title after it, is that intended?  Or should there be a list there?

ghickman (approved):
I'm so glad to see this being tackled, thank you for finding a solution!

---


> the opensafely-databuilder-mkdocs plugin is [tested with 3.7](https://github.com/opensafely-core/mkdocs-opensafely-databuilder/blob/7a1686a5f887e3b35d8dbe5fffd4bcfa066f0599/.github/workflows/main.yml#L30). (Which we should possibly go and change.)

Not for this PR, but wasn't sure of a better place to put it!  The plugin was tested with 3.7 so it was in lock-step with this repo because it needs to be installed into the virtualenv.  So it should be updated to match this one.

Side note: I previously tried to see if it was plausible to move the plugin code into this repo.  Plugins use setuptools entrypoints so if we can work out how to specify a local directory in the requirements spec then it should work.  We'd save a whole repos worth of admin overhead then.

StevenMaude:
> thank you for finding a solution!

It's not the entire way there. But this, along with Becky's recent changes to how the reference page JSON is generated in the Data Builder repository, should be the start of things getting a bit easier.

INSTALL.md
StevenMaude: I tend to agree, although semantic line breaks have benefits in interpreting subsequent changes and diffs. [^1]

I'll remove them here, because:

* there's a mix of styles
* the likely audience (developers) will likely be reading in a terminal or text editor [^2]

[^1]: Semantic line breaks make Serious Content™ read to the eye like slightly whimsical poetry. After some time, I think I'm getting slowly accustomed to at least *reading* this way, but that takes a while. *Writing* is still more awkward, you either have to:
      * make frequent edits to multiple lines when rewriting
      * write without semantic line breaks until you're done, then add them last.

[^2]: The intended audience is an important question. **Where and how are most people going to read this document?**

      If it's as Markdown rendered into another format, semantic line breaks make more sense. If it's most likely going to be read by someone in a text editor or pager, maybe semantic line breaks are more irritating to have than the benefits to reading subsequent diffs.

pyproject.toml
StevenMaude: > Does ~=3.9 work here?

Yes, it should do: https://peps.python.org/pep-0508/#grammar

But I've ditched `[project]` as suggested. We're not building a distribution from this project as such.

And the current requirements files and workflows imply the currently used version.

INSTALL.md
StevenMaude: Done.

INSTALL.md
StevenMaude: I have now removed that "historical" section.

PR #846, opened 2022-07-07
Enumerate all the column types in ehrQL reference

StevenMaude:
Instead of letting the reader work the rest out.

It would possibly be better if we could generate this from the
specification, but will do for now.

### DIFF
--- a/docs/ehrql-reference.md
+++ b/docs/ehrql-reference.md
@@ -24,13 +24,18 @@ Every example here consists of:
 
 1. Headings and subheadings that summarise the feature being demonstrated.
 2. A small example data input table containing entirely fictitious variables and values.
-    * The table has a single-letter name referred to throughout the example;
-      for example, `e` for event-level table
-      or `p` for patient-level table.
-    * The columns of input tables also have single-letter names:
-      these single-letter names refer to the column's data type.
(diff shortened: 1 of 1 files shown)


docs/ehrql-reference.md
```diff
@@ -24,13 +24,18 @@ Every example here consists of:
...
+        * `e` for event-level table
+        * `p` for patient-level table.
+    * The columns of input tables use a name constructed from a single letter with a number
+      to create an identifier — for example, `i1`.
+      The single letter in the identifier refers to the column's data type:
+        * a `b` column contains Boolean values
+        * a `c` column contains electronic health record codes
+          (the codes used in this reference are fictitious, for example: `abc`)
+        * a `d` column contains dates
+        * an `i` column contains integers
```
CarolineMorton, line 37: I think this is clear. 

A thought that I have had though is should we be making this more researcher readable. i.e. a fake study like COPD and infections or something like that so people can understand more clearly what sort of data we are referring to.

CarolineMorton: Ok that makes sense.

CarolineMorton (approved):
Thanks for clarifying

PR #860, opened 2022-07-15
Trigger build check after opening automated update docs PR

rebkwok:
The build-check  workflow is triggerd by PRs, but apparently not by the one opened by the update-databuilder-docs workflow.  I think this is because it's using the GITHUB_TOKEN, which is by default.  So now we manually trigger at the end of the update-databuilder-docs workflow.

I also removed the `team_reviewers: ['data-team']` from the PR step of the update-databuilder-docs workflow; the data-team is in a different organisation so can't be added as a reviewer here.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,7 @@ name: 'build-check'
 # yamllint disable-line rule:truthy
 on:
   pull_request:
+  workflow_dispatch:
 
 jobs:
   build-check:
--- a/.github/workflows/update_databuilder_docs.yml
+++ b/.github/workflows/update_databuilder_docs.yml
@@ -55,7 +55,6 @@ jobs:
               owner,
               repo,
               pull_number: result.data.number,
-              team_reviewers: ['data-team'],
             });
       
       - name: Trigger build-check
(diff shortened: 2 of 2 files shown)


ghickman (approved):

StevenMaude:
> I think this is because it's using the GITHUB_TOKEN, which is by default. 

Yes, [that's correct](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow):

> When you use the repository's `GITHUB_TOKEN` to perform tasks, events triggered by the `GITHUB_TOKEN` will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's `GITHUB_TOKEN`, a new workflow will not run even when the repository contains a workflow configured to run when `push` events occur.

rebkwok:
I think this still isn't triggering the build check; I thought it would because it's using the same mechanism as [databuilder does](https://github.com/opensafely-core/databuilder/blob/main/.github/workflows/update-public-docs.yml#L73) to trigger the update-docs workflow here, but of course databuilder has to use a diffferent token with write access to this repo

PR #866, opened 2022-07-18
Update databuilder install docs and just commands

### LINKED ISSUE #863: Remove databuilder requirement
Now that the public_docs.json is generated automatically by databuilder, we don't need databuilder as a requirement in this repo

rebkwok:
Fixes #863 

Remove just commands to link databuilder and generate docs now that the databuilder JSON is updated automatically.

Also fix a step in the update docs workflow.


iaindillingham (approved):
🚀

PR #871, opened 2022-07-19
Check if branch and PR already exist for databuilder docs updates

rebkwok:
A databuilder docs update workflow can be rerun manually if necessary (e.g. if something went wrong and it didn't complete all its steps); if it's rerunning, we need to check whether the branch being used for commiting the new file already exists (otherwise creating a new branch locall and pushing it will fail).  We also need to check if the PR is already open before creating it, so we don't end up with multiple PRs for the same thing.


iaindillingham (approved):
I'm not a GitHub actions master. However, your description of the problem tallies with your solution 🙂

PR #894, opened 2022-07-29
Clarify public private repos

lucyb:
One for @CLStables to take a look at (and probably others) as part of a larger piece of work.

Update: have asked someone who does co-piloting to take a look before merging

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -169,9 +169,10 @@ account, for developing your own study:
    repository (typically under your own account)
    ![Entering a name, owner and description, when creating a repository from the research template.](images/create-repository-owner-name.png)
 1. Type a name for your repository, and an optional description.
-1. Choose a repository visibility.  This would usually be "Public".
+1. Choose a [repository visibility](repositories.md#repository-access). This would usually be "Public", unless you would like to restrict repository visibility while you're getting started.
+
    ![Entering a description and choosing to make a repository public or private, when creating a repository from the research template.](images/create-repository-public-private.png)
-1. There is an "Include all branches" option: it can be left disabled.
+1. There is an "Include all branches" option: this does **not** need to be selected.
    (You only need the main branch; the other branches are
    work-in-progress changes.)
 1. Click **Create repository from template**
@@ -660,8 +661,7 @@ the new commit.
     1. In the list, tick "public\_repo", which gives
        your Gitpod account "write access to code in public repositories
        and organizations" on GitHub. (There is another permission that
-       you have to enable for access to code in private repositories,
-       though this is not necessary for this walkthrough.)
+       you have to enable for access to code if you've chosen to make your repository private)
 
        ![Opening Gitpod's settings in their code
        editor.](images/gitpod-edit-permissions.png)
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -168,9 +168,8 @@ account, for developing your own study:
 1. Use the **Owner** drop-down menu, and select the account you want to own the
    repository (typically under your own account)
    ![Entering a name, owner and description, when creating a repository from the research template.](images/create-repository-owner-name.png)
-1. Type a name for your repository, and an optional description.
-1. Choose a [repository visibility](repositories.md#repository-access). This would usually be "Public", unless you would like to restrict repository visibility while you're getting started.
-
+1. Type a name for your repository, and an optional description
+1. Choose a [repository visibility](repositories.md#repository-access). This would usually be "Public"
    ![Entering a description and choosing to make a repository public or private, when creating a repository from the research template.](images/create-repository-public-private.png)
 1. There is an "Include all branches" option: this does **not** need to be selected.
    (You only need the main branch; the other branches are
--- a/docs/repositories.md
+++ b/docs/repositories.md
@@ -9,9 +9,11 @@ GitHub is the means by which code in the repository is passed to the server to b
 
 ## Repository access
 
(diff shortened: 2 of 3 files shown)


docs/getting-started.md
```diff
@@ -168,10 +168,10 @@ account, for developing your own study:
 1. Use the **Owner** drop-down menu, and select the account you want to own the
    repository (typically under your own account)
    ![Entering a name, owner and description, when creating a repository from the research template.](images/create-repository-owner-name.png)
-1. Type a name for your repository, and an optional description.
-1. Choose a repository visibility.  This would usually be "Public".
+1. Type a name for your repository, and an optional description
+1. Choose a [repository visibility](repositories.md#repository-access). This would usually be "Public"
    ![Entering a description and choosing to make a repository public or private, when creating a repository from the research template.](images/create-repository-public-private.png)
-1. There is an "Include all branches" option: it can be left disabled.
+1. There is an "Include all branches" option: this does **not** need to be selected.
```
brianmackenna, line 174: Are there any times when any external person would need to use it?

docs/repositories.md
```diff
@@ -47,14 +49,14 @@ Note that if someone else wants to commit to your recently created OpenSAFELY re
...
 ### `README.md`
 
-This file introduces your project, helps people understand the purpose of the repo, and how to navigate and use the code. 
+This file introduces your project, helps people understand the purpose of the repo, and how to navigate and use the code.
 
-Any new repo will include a template README. 
-It should be updated to briefly describe the intended study and to reflect the repo's current status as a _work in progress_. 
+Any new repo will include a template README.
+It should be updated to briefly describe the intended study and to reflect the repo's current status as a _work in progress_.
 The README should also be updated frequently to reflect changes to the repo, and should certainly be up-to-date before publication of any research outputs.
```
brianmackenna, line 56: I wonder should we give examples of one or two repos as best practice?

lucyb: That's a good idea. We have another ticket to [update the README](https://github.com/opensafely/research-template/issues/92) instructions, I'll add the suggestion there. 

The change here is just to remove some whitespace at the end of the lines, otherwise it wouldn't pass the tests.

docs/repositories.md
```diff
@@ -7,11 +7,13 @@ You can download a copy of the repo ("clone"), create a development "branch", ma
...
+In accordance with the [Principles of OpenSAFELY](https://www.opensafely.org/about/#transparency-and-public-logs), we recommend that repositories are made public from the start. A public GitHub repository is visible by anyone on the internet. However, only specified people will have the ability to make changes to it.
 
-Once a project is complete, or retired, the repository will be made public. 
+During the development stage of a project, a repository may be kept private, so that only members of the OpenSAFELY GitHub organisation are able view it.
+
+A repository will need to be made public if it forms part of [a publication](https://www.opensafely.org/policies-for-researchers/#acknowledgment-and-data-sharing--publication-policy). We have a [guide to publishing repositories](publishing-repo.md) that we advise you to read. For unfinished or abandoned research, we strongly encourage that repositories are made public within 12 months after the code is first run against an OpenSAFELY database. You can request that a private repository is made public at any time by contacting [tech support](how-to-get-help.md).
```
LFISHER7, line 16: Is there any need for distinction here? Will we not simply make any repo public 12 months after the first time code is run?

docs/getting-started.md
lucyb: This option is referring to the cloning of the opensafely/research-template repository as part of the setup and they only need to copy over the main branch in that case. It doesn't affect researchers creating branches later on.

docs/permissions.md
```diff
@@ -4,13 +4,9 @@ After the data controller has [approved your project](https://www.opensafely.org
 
 Additionally, the OpenSAFELY framework requires a user to be a member of the `opensafely` Github organisation in order to [run jobs in the secure backends](job-server.md).
 
-## Adding collaborators
 
-While your repository is private, you may need to add new collaborators who are not named in your original application. Any new collaborators have the same information governance obligations as the original study authors, and must therefore sign the same access agreements; please send any such requests to [team@opensafely.org](mailto:team@opensafely.org).
-
-When you add a new repository in the opensafely organisation, it may take up to 15 mintutes for it to be available to select at https://jobs.opensafely.org
-
-## Project Member Management
```
LFISHER7, line 13: I've removed this page as part of #897 and folded all the information into the jobs site page. Would you like to check that this fits with hat you had in mind?

docs/repositories.md
lucyb: It may be made public sooner if for some reason the work is published earlier than 12 months. People may also choose to make it public before then.

Technically, we're asking people to agree to make it public rather than forcing it to be public too.

LFISHER7: This seems to be more permissive than what's written [here](https://www.opensafely.org/about/#transparency-and-public-logs).

I'm not sure exactly what is in any agreements signed in an application, but the below suggests this is made clear.

> OpenSAFELY requires all researchers to archive and publish their analytic code: this is the only way they are allowed to run code against real data

>

docs/permissions.md
lucyb: Fab, yes that's perfect. I'll not merge any changes to that file then to avoid conflicts.

The only thing I'd suggest is that you link to the https://docs.opensafely.org/publishing-repo/ page in the `Publishing outputs` section otherwise you risk repeating information.

docs/repositories.md
sebbacon: Links need fixing, sorry 

Also just to check, they do have to _ask an admin_ to make it public, right? Amir has been saying to people they can do it themselves, following approval. If he needs correctly, please let him know in Slack. If not, the above needs tweaking

ghickman (approved):

PR #895, opened 2022-08-01
Add rounding instructions to disclosure docs

LFISHER7:
This could be expanded on in the future, but provides at least some instruction for rounding.

### DIFF
--- a/docs/releasing-files.md
+++ b/docs/releasing-files.md
@@ -52,7 +52,6 @@ Primary disclosure describes the situation where confidential data can be obtain
 |           | 41-50        | 15                | 90             |
 |           | 51+          | 25                | 85             |
 | **Total** |              | 51                | 276            |
-|           |              |                   |                |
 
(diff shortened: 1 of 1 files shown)


StevenMaude:
suggestion: Would it be useful to have a concrete example with fictitious data? This could show tables *before* and *after* the redaction and rounding process.

This might help illuminate the details, in case a reader either misinterprets the instructions or is otherwise unclear on the meaning.

It's possible to add [simple tables in the Markdown](https://squidfunk.github.io/mkdocs-material/reference/data-tables/) without too much work.

StevenMaude:
Otherwise, the text looks fine. But better for someone who knows our disclosure controls well to approve it :mag:

LFISHER7:
Thanks @StevenMaude , that's a good suggestion, I'll get something added

Jongmassey (approved):
Both the text and example are excellent, and I look forward to referring researchers to this in future!

PR #885, opened 2022-07-21
Rebuild databuilder dependencies for Python 3.9

StevenMaude:
Fixes #882.

In #832, we made everything run with Python 3.9.

This set of requirements was missed.

This also upgrades the `databuilder` version.

PR #900, opened 2022-08-04
Wrangle black version for consistency

StevenMaude:
This was broken.

pre-commit would try to install Black 21.12b0, but with click 8 and this didn't work.

See #885 and commit message 6bf6a7bdbbdd805b624b2cf83deb5f7d78efbdcf for more on why we're in this situation.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,6 +3,8 @@ default_language_version:
 
 repos:
   - repo: https://github.com/psf/black
+    # Pinned temporarily for consistency with the databuilder directory's
+    # requirements; see #886.
     rev: 21.12b0
     hooks:
     - id: black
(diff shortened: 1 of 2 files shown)


ghickman (commented):
Oh this is so painful that snex is holding us to the broken version 🤦‍♂️

Good catch though!

ghickman (approved):

StevenMaude:
@ghickman: Yeah, it's a little painful; I think the black requirement of click was actually set to >=v7.x, so it was pulling an incompatible version.

We could fork snex, but we can manage for now as is. It's just a little unfortunate snex uses click: most of the snex dependencies are independent of everything in this otherwise.

We also have a draft ticket on the data team backlog board to review how the Data Builder documentation is integrated.

PR #902, opened 2022-08-08
Alexjs linter improvements

lucyb:
Use of words like "just", "simple" or "easy" may be insensitive.


iaindillingham (approved):
Looks good to me. "Just" is an expensive word, so we should seek to remove it from the documentation 🙂

Issue #834, opened 2022-06-30
Expand on the current tutorial to demonstrate ehrQL concepts


StevenMaude:
# Goal

The existing ehrQL reference, based on the specification tests, is good at describing behaviours, but does not make an attempt to explain them.

The current tutorial was written as a placeholder. We should now extend the existing placeholder tutorial.

# Tasks

This requires:

* determining the core ehrQL concepts a user needs to understand, and those that can be ruled out
* developing a simple example dataset (details in #833) to use in the tutorial
* writing simple ehrQL examples that demonstrate those concepts
* writing surrounding explanatory text
* the surrounding text on `https://docs.opensafely.org/ehrql-intro` modifying, as the tutorial will change; we can leave the simple example as is.

This could be completed independently of fully integrating the tutorial into this documentation (#835). It would be sufficient, though not ideal, to include code initially as static Markdown content [^1].

[^1]: A longer-term goal should be to require that any working code examples in newly added documentation is suitably tested, just as we typically expect newly added software code is suitably tested.

StevenMaude:
## Potential ehrQL concepts to cover

* The current ehrQL interface is based on Python. 
* `Frames`, `Series`
  * It may be worth mentioning the insipiration of terminology from pandas.
* … :thought_balloon:

StevenMaude:
Feedback collected from various comments:

* Perhaps to use a full `project.yaml` instead of just a dataset definition.
* Include learning objectives and small assessments.
* Cut down on admonitions where possible.
* Consider use of collapsible blocks for admonitions.
* Review what needs to be include; do we need to explain everything?

inglesp:
> Review what needs to be include; do we need to explain everything?

Is this at the level of reviewing which databuilder concepts need to be explained, or which extra stuff (eg black/isort/flake8) need to be explained?  (Or both?)

StevenMaude:
> Is this at the level of reviewing which databuilder concepts need to be explained, or which extra stuff (eg black/isort/flake8) need to be explained? (Or both?)

Initially, I was thinking more "extra stuff" because that's primarily what I took from a comment that was made.

But I think it relates to everything. That should become clearer when people work through.

StevenMaude:
I've combined the [feedback above](https://github.com/opensafely/documentation/issues/834#issuecomment-1245240118) into a checklist in #987.

PR #912, opened 2022-08-10
Start expanding ehrQL tutorial: add first dataset definitions

StevenMaude:
Relates to #834.

This may just be an initial addition, instead of a complete document.

We'll merge based on having a reasonably complete explanation of the first dataset definitions.


CarolineMorton:
*/data-builder-intro*
- This is great as it is well structured
- Is it definitely Data Builder or Databuilder (I have been using the latter but I think this might be wrong?)
- ```**We no longer support specifying `latest` as a version.**```  - is important but currently formats as a heading (or looks like a heading rather). 
- ```For instance, one backend might use Microsoft SQL Server and another Databricks.``` => add after, `This means that they need the query to be written in a different way to generate the same dataset. This conversion into a compatible query is done by the query language and does not need to be handled by the researcher. This has the advantage of meaning the same dataset defintion can be used in more than one backend without the need to write two dataset definitions`

*/ehrql-intro*
- This needs more information about what ehrQL is and perhaps some examples of the expressiveness. Perhaps something like the contents (not the fanciness of https://prql-lang.org/). I like the principles here (we could add ehrQL principles) and the simple showcase. 

*ehr tutorial*
- necessary requirements -  querying databases or other data stores. Not sure this is true. Could we soften this?
- Add an example table of what an output might look like

StevenMaude:
@CarolineMorton:

Thanks for looking end-to-end; appreciate it as I haven't in some time, and a lot of the other pages were written a few months back! :mag:

Some of the pages do need a re-visit and possible pruning before we add all the Data Builder related information to the main table of contents.

There's possibly some repetition. A little bit, say a line or two explaining something is OK, provided it points to the canonical explanation.

> _/data-builder-intro_
> 
>     * This is great as it is well structured
> 
>     * Is it definitely Data Builder or Databuilder (I have been using the latter but I think this might be wrong?)

As far as I recall, this is the convention I've used: "Data Builder" as a software name, `databuilder` as the CLI tool.
 
>     * `` **We no longer support specifying `latest` as a version.** ``  - is important but currently formats as a heading (or looks like a heading rather).

I've just fixed this.

>     * `For instance, one backend might use Microsoft SQL Server and another Databricks.` => add after, `This means that they need the query to be written in a different way to generate the same dataset. This conversion into a compatible query is done by the query language and does not need to be handled by the researcher. This has the advantage of meaning the same dataset defintion can be used in more than one backend without the need to write two dataset definitions`

I've modified this section, mostly as you've suggested with some edits.

> 
> _/ehrql-intro_
> 
>     * This needs more information about what ehrQL is and perhaps some examples of the expressiveness. Perhaps something like the contents (not the fanciness of https://prql-lang.org/). I like the principles here (we could add ehrQL principles) and the simple showcase.
> 

This will need a bit more thinking about. Please remind me if I forget to return to this.

Separately, I feel like a comparison either between SQL and ehrQL, and/or study definition and ehrQL might be useful here, or maybe in a separate page. It might be a little offputting, if you're *not* familiar with them.

> _ehr tutorial_
> 
>     * necessary requirements -  querying databases or other data stores. Not sure this is true. Could we soften this?

I think you're right: it might deter people, even if they've done lots of pandas/R/Stata.

I'll just remove that line.

>     * Add an example table of what an output might look like

Could you be a little more specific here? I'm intending to include outputs, for each dataset definitions, and already include some.

inglesp (approved):
This is a really solid start, and the `justfile` machinery will be really useful for adding to this.

It's hard for me to know whether the content is at the right level (but it looks to be in the right ballpark!) and so we should think about trying this out on users who are new to databuilder, and continuing to get feedback as you make changes.

PR #897, opened 2022-08-02
Update jobs docs

LFISHER7:
This extends the existing jobs site documentation to include information on the projects/workspaces/jobs hierarchy, the different types of outputs types (resolves #600) and viewing and publishing outputs.

There are a few details that I need to check are correct, so leaving in draft for now

Issue #764, opened 2022-06-01
Re-enable `--strict` in site build


StevenMaude:
Following #763.

StevenMaude:
Related issue on configuring these warnings: mkdocstrings/mkdocstrings#437

StevenMaude:
I think it's finally possible to disable all the mkdocstrings warnings now.

PR #919, opened 2022-08-18
Fix internal links to job-server and permissions pages

StevenMaude:
Following #897 and #918.


StevenMaude:
It's easy to miss these due to #764: mkdocstrings spams several hundred lines of logs that (in our case) is unhelpful.

rebkwok (approved):
Not necessarily for this PR, but the new jobs docs refer to job server as "the jobs site" now.  There are several references here to "job server" still that we should update.

StevenMaude:
> There are several references here to "job server" still that we should update.

Done; thanks for highlighting this!

PR #920, opened 2022-08-18
Add starting work for second batch of dataset definitions

StevenMaude:
As well as the newer content, there are some significant changes here:

* Rework the dataset definitions slightly to accommodate the Data Builder refactor.
* Split the tutorial into multiple pages. The starting page is now at `/ehrql-new-tutorial-intro/` and the others are linked from there. The final page is a placeholder which contains headings for subsequent pages that aren't yet written.

***

The actual dataset definitions in question are not yet complete, but I would still like to merge as an ongoing work-in-progress. They all have big prominent warnings stating they are unfinished.

This means:
* there's not one big PR to review
* the current state is accessible from the main deployment
* we're not waiting to merge until the perfectionist in me is happy; there is still ongoing change in Data Builder which means these pages are subject to change anyway

As @inglesp and I have discussed, the tidiness of the commit history here is less important than in a code-focused repository: there may be subsequent revisions to this content, but that's OK.


CarolineMorton:
Hi @StevenMaude 

This is great. I don't have much to add at this stage except maybe breaking this into a series of tutorials, with learning objectives at the top of each one.

ghickman (approved):

docs/ehrql-new-tutorial-intro.md
```diff
@@ -0,0 +1,297 @@
...
+!!! tip
+    There is no requirement for approval or access to OpenSAFELY to follow this tutorial.
+    This tutorial uses artificial data throughout.
+
+## What this tutorial covers: writing and running ehrQL queries with Data Builder
+
+ehrQL is a query language for electronic health records.
+The [introduction to ehrQL](ehrql-intro.md) has more information.
+
+Data Builder is a program that takes ehrQL queries,
```
sebbacon, line 44: I understood this as "a programme of work" which created a pause in my parsing.

Perhaps "is a software component" or just "is software"?

sebbacon: Separately - I find "Data Builder" hard to parse. This is just a note that perhaps we should revisit naming of components one more time when user testing things like names and syntax with real people.

sebbacon:
I really like the way these examples are shaping up.

My one bit of general feedback is that it feels like there's a lot of tips, warnings, code examples, and headings, all of which are accurate and serve a purpose, but which (in combination with the current styling, lots of vertical space pushing content down the page, headings, bullet points) contribute to a sense of a lot of noisy "chrome".  

On the assumption this is a "getting started" replacement guide, I feel like we want something as close to one page as possible for each step, which gives the user a "tada!" feeling. The objective for the first page, "run a dataset", seems right, but there's a lot of scrolling and reading to get there - e.g.:

image
Some options (if others agree - I'm just one user) (a) think about the design a bit so the narrative flow is more compressed (e.g. can we pull tips out to the side?) (b) move out the exceptions and tips and qualifications to other pages and link to them, rather than inline them; or move them to subsequent learning objectives (c) maybe just cull some of the detail, e.g. what if we _don't_ explain what dots do in Python? Or what `black` is? (d) what would it look like if we attempted to restrict ourselves to one learning objective per page?

Hope that makes sense, happy to brainstorm in a call if/when useful!

StevenMaude:
Thanks to everyone who's taken a look and shared their thoughts. It's useful to know what others think when I'm staring very closely at this.

@sebbacon: Yes, I agree there's too heavy use of admonitions. My "todos" liberally sprinkled throughout don't help, though these are useful for now to both track and highlight incomplete sections. Those should eventually go :crossed_fingers:

It's possible to make admonitions [collapsible](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#collapsible-blocks). This might help with the clutter and make them feel more like an optional section.

As has been pointed out, some of those additional notes might be less useful or a distraction at this stage. This goes particularly for the notes on Python tooling as Simon pointed out. I'll revisit what needs to be in there.

I won't make most of these changes here as I'd prefer to merge sooner, but I've [noted them](https://github.com/opensafely/documentation/issues/834#issuecomment-1245240118) in the tracking issue.

PR #922, opened 2022-08-19
Update study-def-dates.md

andrewscolm:
Add specific instructions for working with dates using `patients.satisfying`.

### DIFF
--- a/docs/study-def-dates.md
+++ b/docs/study-def-dates.md
@@ -22,6 +22,25 @@ You should be aware that events can be recorded in clinical systems without a da
  
 You should take this into account in your analysis; for example, you might want to discard blood pressure values that are in the future, but you might want to keep an ethnicity category which was recorded with no date, as these are less likely to change over time. 
 
+### Using dates with patients.satisfying
+
+You should be aware that `patients.satisfying` does not treat `1900-01-01` as equivalent to `NULL`. In order to exclude impossible dates you might want to add
+checks such as
+
+```
+DIAB_GROUP = patients.satisfying(
+                """
+                (
+                    DIAB_DAT
+                    AND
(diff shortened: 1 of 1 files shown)


docs/study-def-dates.md
```diff
@@ -22,6 +22,25 @@ You should be aware that events can be recorded in clinical systems without a da
  
 You should take this into account in your analysis; for example, you might want to discard blood pressure values that are in the future, but you might want to keep an ethnicity category which was recorded with no date, as these are less likely to change over time. 
 
+### Using dates with patients.satisfying
+
+You should be aware that `patients.satisfying` does not treat `1900-01-01` as equivalent to `NULL`. In order to exclude impossible dates you might want to add
```
ccunningham101, line 27: I think this is definitely useful! Maybe instead of "in order to exclude impossible dates", it could be "in order to exclude one impossible date such as 1900-01-01" as to highlight that there could be other impossible dates and this may not cover it all?

LisaHopcroft: Agreed - definitely useful but @ccunningham101 has evidence that there are other impossible dates.

Maybe change to: "in order to exclude this and other impossible date (e.g., "9999-12-31", ...) you might want to add"...?

LisaHopcroft: While you're in here (!), there are dates in the previous section that are not inside backticks (ie "9999-12-31" and not `9999-12-31`). I think they would stand out more in the text if they were formatted the same way as you have formatted them in this section.

LisaHopcroft (commented):
Didn't really mean to press the "start review button" (!). Only two small comments.

ccunningham101 (approved):

PR #928, opened 2022-08-31
Add new section to opensafely cli docs about resource management

bloodearnest:


bloodearnest:
@sebbacon added some [comments](https://github.com/opensafely-core/opensafely-cli/pull/137#issuecomment-1232863613) suggesting cross linking from memory-efficent-working.md to this new resources concurrency section.

There's not an obvious place on that page to do that from, as which is all about code changes.

I think the motivation is that that page is read more frequently? I'm not sure this is true - there are 0 crosslinks to `memory-efficient-working.md` currently in the code base, where as there are 8 crosslinks to `opensafely-cli.md`.

sebbacon:
Motivation: 

I think memory issues are one of those wicked snafus that can cause a lot of pain; I think this new feature is likely to be very helpful as a way of setting context for users to starting thinking "how much memory do I have available".

Hypothesis: I don't think people reach for the documentation of a toolbelt command when they are having memory issues, I think they start search for the word "memory" or such.

So linking _from_ pages about memory _to_ the documentation will be useful.

However I think @StevenMaude has the best overview of how people use the docs, which is why I tagged him.

StevenMaude:
I think there's no harm, and possible benefit, in linking thematically sections together.

I also agree with Simon that there's nothing in the "Writing performant code" page currently where a link would fit. So I'd suggest just adding a short admonition to the top of the ["Writing performant code" page](https://docs.opensafely.org/memory-efficient-working/) along the lines of:

```
This section provides advice on improving the performance of your scripted actions … the principles can also apply to other languages (such as Stata).

!!! note
    When developing your code, there are options in the OpenSAFELY CLI you can use to control the [computing resources used](opensafely-cli.md#managing-resources).
    These options may be useful in conjunction with optimising your code following the guidance here.
```

StevenMaude:
> I think the motivation is that that page is read more frequently? I'm not sure this is true

In principle, we could find this from Plausible's dashboard. Though I think the CLI page will be read more though.

StevenMaude:
Otherwise, this looks fine. The only nit is lowercase `docker` when it should be the name of the software, but not really a blocker.

bloodearnest:
Have added the note and fixed Docker capitalisation.

Can I get a +1 from someone so I can land?

ghickman (approved):

PR #929, opened 2022-08-31
Skip publish step for Dependabot

StevenMaude:
Dependabot does not have access to secrets, so that malicious
dependencies cannot exfiltrate those secrets.

Because of this, the publish step fails for Dependabot PRs as the
Cloudflare secrets are not available.

We could consider a more sophisticated separation of build and publish
in future.


StevenMaude:
This is a sticking plaster fix. I've opened #930 to consider improving this in future.

tomodwyer (approved):

PR #931, opened 2022-08-31
Include last revision date for each page

### LINKED ISSUE #762: Add a date or commit ID somewhere to published HTML

StevenMaude:
Fixes #762.

### DIFF
--- a/.github/workflows/pages-deployment.yml
+++ b/.github/workflows/pages-deployment.yml
@@ -10,6 +10,8 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: true
+          # Fetch entire repository history to determine last revision dates of pages.
+          fetch-depth: 0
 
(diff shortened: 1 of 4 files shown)


StevenMaude:
Mixed feelings about this: it does work now. But the plugin simply checks the last commit date for a given source file: if that source file has includes that have been more recently updated, then the date is unchanged. So it's potentially slightly misleading.

StevenMaude:
Closed this as it's not quite semantically correct, as things stand.

PR #939, opened 2022-09-06
Document the `--param` argument to cohortextractor

evansd:
As added in:
https://github.com/opensafely-core/cohort-extractor/pull/860

Preview at:
https://evansd-parameterised-study-d.opensafely-docs.pages.dev/study-def/#parameterised-study-definitions

### DIFF
--- a/docs/study-def.md
+++ b/docs/study-def.md
@@ -246,7 +246,9 @@ actions:
 
 ### Parameterised study definitions
 
-If you want to have multiple study definitions which are structurally similar but differ in the values of a few parameters then you can avoid having to create multiple study definition files by using the `--param` argument. For example, below we create two actions which use the same study definition but supply different values for `my_param`:
+Parameterised study definitions allow you to create multiple study definitions that are structurally similar but differ in the values of a few parameters. This removes the need to create multiple similar study definition files.
+
+First, set the relevant parameters in the action definition using the `--param` argument. For example, below we create two actions which use the same study definition but supply different values for `my_param`:
 
 ```yaml
 version: "3.0"
--- a/docs/study-def.md
+++ b/docs/study-def.md
@@ -312,7 +312,7 @@ generate_cohort --param my_param='Value with spaces' ...
 ```
 
 ### Identical study definitions with different index dates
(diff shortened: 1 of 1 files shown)


docs/study-def.md
```diff
@@ -243,12 +243,77 @@ actions:
         cohort: output/input_asthma.csv
 ```
 
+
+### Parameterised study definitions
+
+If you want to have multiple study definitions which are structurally similar but differ in the values of a few parameters then you can avoid having to create multiple study definition files by using the `--param` argument. For example, below we create two actions which use the same study definition but supply different values for `my_param`:
```
StevenMaude, line 249: suggestion: This sentence is quite long. It is also critical to understanding the idea here.

Perhaps consider breaking it up into something like:

```suggestion
Parameterised study definitions allow you to create multiple study definitions that are structurally similar but differ in the values of a few parameters. This removes the need to create multiple similar study definition files.

First, set the relevant parameters in the action definition using the `--param` argument. For example, below we create two actions which use the same study definition but supply different values for `my_param`:
```

docs/study-def.md
```diff
@@ -243,12 +243,77 @@ actions:
...
+```
+
+If parameter values contain spaces you will need to quote them:
+```
+generate_cohort --param my_param='Value with spaces' ...
+```
+
 ### Identical study definitions with different index dates
-If the only difference between two study definitions is the index date, then you can avoid having to create two separate study definition files as follows.
+If the only element of a study you wish to parameterise is the index date then, rather than using the `--param` argument, it might be easier to use the `--index-date-range` argument from the [measures](measures.md#extract-the-data) feature, which can be given a single date argument instead of a range.
```
StevenMaude, line 315: suggestion: I'd personally prefer trying to break up sentences wherever possible. Maybe like:
 
```suggestion
To parameterise index dates, there is a simpler alternative than using `--param`. Use the `--index-date-range` argument from the [measures](measures.md#extract-the-data) feature. `--index-date-range` can be given a single date, instead of a range.
```

StevenMaude (approved):

StevenMaude:
Looks fine. I just added small, optional editorial notes that might help a reader follow the text more easily.

evansd:
Thanks @StevenMaude, both good suggestions which I've incorporated.

PR #836, opened 2022-06-30
feat: document first honeycomb related jobs link

madwort:
https://github.com/opensafely-core/job-server/pull/1898

PR #948, opened 2022-09-09
feat: move monitoring docs to team manual

madwort:


docs/timings-monitoring.md
```diff
@@ -1,12 +1,3 @@
...
-    These notes are a work-in-progress.
-    This functionality is in limited release and subject to change.
-
-## Audience
-
-This documentation is aimed at researchers who want feedback on the performance of their job requests, as well as the performance context in which their job requests happen (i.e. was the server busy or quiet at the time).
-
 ## Timings
 
 Basic timing information is available from job and job request detail pages, for example:
```
lucyb, line 3: Is it worth moving this into another part of the manual or removing the page altogether for the time-being? I wouldn't expect anyone to have linked to it and it's not currently providing the end user with much actionable information.

madwort: yeah fair enough, will remove

lucyb (approved):

StevenMaude:
If this is merged, it looks like #836 should be closed, without merging; is that correct?

lucyb:
> If this is merged, it looks like #836 should be closed, without merging; is that correct?

That's correct. I'll merge this now, since Tom is away.

PR #952, opened 2022-09-13
Add `project.yaml` explainer for Data Builder

StevenMaude:
This is not completely finished. However, to do so, requires getting the
`project.yaml` working, which may not be possible yet.

In future, we may prefer to integrate this information earlier, and
replace the existing examples that run with Data Builder directly, with
entire OpenSAFELY projects: one for each dataset definition.

For now, this at least makes the connection between Data Builder and
OpenSAFELY `project.yaml` explicit.

### DIFF
--- a/databuilder/ehrql-tutorial-examples/project.yaml
+++ b/databuilder/ehrql-tutorial-examples/project.yaml
@@ -1,0 +1,11 @@
+version: "3.0"
+
+expectations:
+  population_size: 3000
+
+actions:
+  extract_minimal_cohort:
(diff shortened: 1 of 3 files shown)


databuilder/ehrql-tutorial-examples/project.yaml
```diff
@@ -0,0 +1,11 @@
+version: "3.0"
+
+expectations:
```
StevenMaude, line 3: We also have to have `expectations` in there, which isn't great. But the YAML doesn't work yet anyway.

CarolineMorton (approved):
I think this all looks good.

PR #955, opened 2022-09-14
Fix typo in study definition tricks example

rebkwok:


docs/study-def-tricks.md
```diff
@@ -170,9 +170,9 @@ study = StudyDefinition(
 
 	**with_these_clinical_events_date_X(
 		name = "probable_covid_date",
-		codelist = probable_codes,
+		codelist=probable_codes,
```
CarolineMorton, line 173: Minor: It feels like all the arguments should have no spaces around the `=` if we have changed a index_date and codelist. 

Otherwise great.

CarolineMorton (approved):

docs/study-def-tricks.md
rebkwok: they should not - I fixed the rest but missed the `name` one.  I'll fix that one too, thanks

rebkwok: I didn't fix all the others, apparently, but I have now :)

PR #964, opened 2022-09-20
Update Stats NZ PDF URL to `archive.org`

StevenMaude:
The PDF has disappeared.

There is a fourth edition at this official-ish URL:

https://statsnz.contentdm.oclc.org/digital/collection/p20045coll4/id/236/rec/1

but it's older than the version originally linked.


LFISHER7 (approved):
Thanks for picking this up @StevenMaude! The new link looks good, the earlier version still includes all of the important info needed

StevenMaude:
@LFISHER7: just to double check; by "new link", do you mean "we should switch the link to the Stats NZ fourth edition PDF on `oclc.org`"?

LFISHER7:
Sorry @StevenMaude, that was ambiguous! The archive version looks good

PR #985, opened 2022-10-03
ehrQL tutorial comments from Milan

milanwiedemann:
- Add missing quotation mark to run command


StevenMaude:
This is still a valid fix — I checked today — but there is now a merge conflict because the file got moved in the meantime.

We should make it in the Data Builder repository, once the migration in https://github.com/opensafely/documentation/pull/1095 is complete. The file being modified will end up there.

So I'll leave it open until we make that change.

StevenMaude:
Closing this as it's part of opensafely-core/databuilder#963.

PR #1001, opened 2022-10-20
Implement more tutorial refinements

StevenMaude:
This PR makes several improvements to the tutorial content. Primarily these are around installation. We can now get a reader to install a current version of OpenSAFELY CLI and use a  public released Docker `databuilder` image to run the examples.

`opensafely exec` is also referred to for now, with equal prominence to `opensafely run`. In the near future, it is likely that we will favour `opensafely exec`. Rather than wait longer to get these improvements in, it is preferable to merge soon, and then later make those changes.

* Format question prompts consistently
* Switch to using a main release version of Data Builder based on a tag
* Add a `project.yaml`
* Check the dataset definitions run with OpenSAFELY CLI in our CI workflow
* Promote OpenSAFELY CLI as the main route to installing Data Builder in the tutorial, with both `opensafely run` and `opensafely exec` as options

# Tasks

- [x] Update documentation to favour Docker-based `opensafely run` over Python-installed `databuilder generate-dataset`
- [x] Update requirements files
- [x] Add `just` recipe to build outputs with `opensafely`
- [x] Test both ways of generating outputs (OpenSAFELY CLI and Python) in CI
- [x] Review what's the best way to check that outputs are consistent.
- [x] Use `opensafely exec` — going to defer [pending further improvements]( https://github.com/opensafely-core/opensafely-cli/issues/149#issuecomment-1293513246) — but will add a note that it's an option.


docs/ehrql-new-tutorial-intro.md
```diff
@@ -228,14 +200,51 @@ and query that data instead of a backend.
...
+    before doing so.
+
+#### Using the OpenSAFELY CLI to run a dataset definition via `opensafely exec`
+
+!!! warning
+    This requires:
+    * OpenSAFELY CLI version v1.32.1 or later
+      (run `opensafely upgrade` to upgrade)
+    * Data Builder Docker image v0.57.0 or later
+      (run `opensafely pull` to update if you already have
```
benbc, line 236: This text is truncated.

databuilder/ehrql-tutorial-examples/project.yaml
```diff
@@ -4,8 +4,56 @@ expectations:
   population_size: 3000
 
 actions:
-  extract_minimal_population:
-    run: databuilder:v0 generate-dataset "./1a_minimal_dataset_definition.py" --dummy-tables "example-data/minimal" --output "outputs/1a_minimal_dataset_definition.csv"
+  extract_1a_minimal_population:
+    run: databuilder:v0.57.0 generate-dataset "./1a_minimal_dataset_definition.py" --dummy-tables "example-data/minimal" --output "outputs/1a_minimal_dataset_definition.csv"
```
CarolineMorton, line 8: Should this not be `databuilder:dev`

CarolineMorton (approved):
@StevenMaude This looks great. Thank you for incorporating the feedback and decisions that we made together when we went through this. 

I have managed to get `opensafely run` to work which feels like massive achievement so thanks for the clear instructions. 

I would suggest we merge quickly and then iterate.

databuilder/ehrql-tutorial-examples/project.yaml
StevenMaude: Perhaps! But this is slightly trickier than it looks.

It's deliberately tied to the Python `requirements.dev.txt` for now, to ensure consistency, when running the examples.

1. It's not possible to just specify an equivalent "latest v0 version" for those requirements: we specify hash checks there, which requires that we know what the hash is, and have to keep that updated.
2. If we instead specify a vague tag like `v0` or `dev`, then I assume that this might try and run with an older `v0` or `dev` image, if someone has an older `v0` image already on their machine and doesn't update. If we're at v1 and ensuring no breaking changes, then it might be more reasonable to pin to v1.

In the meantime, we could add an automated workflow in future to keep the images in sync with the requirements specified for Python.

docs/ehrql-new-tutorial-intro.md
StevenMaude: Thanks; now fixed, and updated since that command wouldn't work anyway right now.

Issue #804, opened 2022-06-15
Update screenshots regularly


StevenMaude:
# Summary

As mentioned by @ghickman on Slack, we could use something like [shot-scraper](https://github.com/simonw/shot-scraper) to generate screenshots.

This ideally would be automated or, failing that, semi-automated and run by a developer locally.

## Benefits

Inspired by https://twitter.com/aaronbassett/status/1536708194160070658:

* For users, this keeps screenshots up-to-date with the current appearance of web services.
* For developers, these act as simple additional checks, warning us when web services have changed substantially, either when:
  * the layout has changed somewhat, and a new screenshot does not represent what was intended
  * the layout has changed enough that this automation fails

## Questions/ideas

* What's the best way of automating login to external services where necessary?
  * Relevant issue for shot-scraper: https://github.com/simonw/shot-scraper/issues/30
  * Can we configure 2FA for such services to secure test accounts? This seems a little tricky.
* Would some services block GitHub Actions runners? I've seen this when automating link checks.
* For services that we develop, we should probably run them with a GitHub Actions runner to screenshot, instead of trying to login into the production site.

StevenMaude:
Another problem: annotating screenshots where they have extra labelling. Example: https://github.com/opensafely/documentation/blob/bd84ebf8bc0dd0b9f86fa782bba2fbb94b9b4538/docs/images/adding-codelist-id-tag.png

The current route via shot-scraper is clunky, though there's an [open issue](https://github.com/simonw/shot-scraper/issues/9) to improve it.

PR #1006, opened 2022-10-25
Document various bits for wrapping up a project

ghickman:
When wrapping up a project we need users to do a few things.  This has changed with the move away readmes being the place to track project updates, so this edits a few pages where we talk about those bits or repos.

Fixes: opensafely-core/job-server#2222


docs/publishing-repo.md
```diff
@@ -1,20 +1,58 @@
...
+## README
+As of the 24th October 2022 we have changed how we managed repo READMEs.
+You can ignore this section if you created your repo after that date.
+
+If you created your repo before this date then you will have the older README format, which expected you to update the file as your project progressed.
+We have now moved to keeping that information in your project on the Jobs site.
+
+The [current README template](https://github.com/opensafely/research-template/blob/main/README.md) has the updated version for new repos.
+You can update your repo by replacing the relevant sections in your own README (typically just the first section) with the content from the template.
+You will also need to replace the placeholder strings in the "View on OpenSAFELY" URL:
```
StevenMaude, line 20: question: Does this not correctly get updated? There is [some kind of script](https://github.com/opensafely/research-template/blob/7061202277ea9bc70837894db066673f8b03c637/.github/workflows/setup.yaml which is supposed to do this, I think.

docs/publishing-repo.md
```diff
@@ -1,20 +1,58 @@
...
+A Release is a coherent snapshot of the code as it existed at a given point in time, along with a title and a description.
+By default, releases are displayed on your repo home page.
+For example, you might make separate releases of the code to accompany a preprint and the final versions of the paper.
+
+Consider [obtaining a DOI for your releases](https://guides.github.com/activities/citable-code/), and using that for referencing your code.
+
+
+## Make your repo public
+If your repository is *private*, ask [tech support](how-to-get-help.md) to make it *public* before the associated paper or output is published.
+There's more information about what this means [here](repositories.md#repository-access).
```
StevenMaude, line 55: suggestion: Drop the [use of "here" in links](https://webaim.org/techniques/hypertext/link_text#uninformative). For example:

```suggestion
The [repositories page](repositories.md#repository-access) has more information on controlling repository access.
```

docs/jobs-site.md
```diff
@@ -140,11 +140,25 @@ Once reviewed, approved and released, your requested files will be available to
...
 
 Once approved, your released outputs will be “published” and viewable from the published outputs in the `Published outputs` section of your `Releases`. This is accessible by everyone, even those without a login.
 
+
+## Updating project status
+When you start a new project on OpenSAFELY it will have the _Ongoing_ status, but as your work progresses you'll want to update this status.
+
+If you are a member of a project you will see an `Edit` button near the top of your project page:
+
+![Project edit button](./images/project-edit-button.jpg)
```
StevenMaude, line 157: thought: The text in `[ ]` is the alt tag. Should we be using more descriptive alt tags to improve accessibility? (It's not necessarily to fix in this particular PR, more a general thought.)

StevenMaude:
nit: Potentially, though I'm not certain, the two JPEG screenshots _might_ compress better and be lossless as a bonus, if you took PNG screenshots. The images are mostly solid colour.

docs/publishing-repo.md
```diff
@@ -1,20 +1,58 @@
...
+* Are there old issues that can be closed? Some issues may have been fixed, or may be no longer relevant.
+
+
+## GitHub Release
+* Consider making a GitHub _Release_, and linking to that ([instructions here](https://docs.github.com/en/github/administering-a-repository/releasing-projects-on-github/managing-releases-in-a-repository)).
+A Release is a coherent snapshot of the code as it existed at a given point in time, along with a title and a description.
+By default, releases are displayed on your repo home page.
+For example, you might make separate releases of the code to accompany a preprint and the final versions of the paper.
+
+Consider [obtaining a DOI for your releases](https://guides.github.com/activities/citable-code/), and using that for referencing your code.
```
StevenMaude, line 50: question: Does this also want to be a bullet point?

StevenMaude: thought: So @ghickman has tried to make the list clearer for what is required; which is really good.

Another improvement might be using the [task list feature](https://squidfunk.github.io/mkdocs-material/reference/lists/#using-task-lists) of Material for MkDocs might be better to highlight the things that are specifically to check? Not sure, I don't think we use that feature anywhere already.

StevenMaude (approved):

docs/publishing-repo.md
ghickman: It does for _new_ repos, this paragraph is attempting to explain how an existing repo could update their README to the new version.  Any how I could better sign post that it's for repos made before the new template went live?

ghickman:
@StevenMaude – I've replace the jpegs (one was actually a png masquerading, no idea why my screenshot app gave me that 🤷‍♂️) with pngs, and optimised them with `optipng -o7 <path>`.

ghickman:
Thought: we should pick a standard for screenshots so we can add a bit of tooling to make it easier.  There's an argument that jpegs can be otpimised to add progressive enhancement into the image file itself, but I've no particular opinion either way.

Relatedly: I'd love to automate these screenshots somehow, particularly with a UI rebuild coming along.  It's almost certainly not worth it at our scale, but still!

docs/publishing-repo.md
ghickman: The bullet point at the start was a vestige of the old page layout, fixed and rebased into the original commit.

StevenMaude:
> Thought: we should pick a standard for screenshots so we can add a bit of tooling to make it easier. There's an argument that jpegs can be otpimised to add progressive enhancement into the image file itself, but I've no particular opinion either way.

I had a quick look and couldn't find any MkDocs tooling for this, but maybe I'm not looking in the right place. It would be good if we didn't have to think too much about this, when adding new images.
 
> Relatedly: I'd love to automate these screenshots somehow, particularly with a UI rebuild coming along. It's almost certainly not worth it at our scale, but still!

Yes, we talked about this previously (#804) :grin: and I think it would be good. There are complications that some images are of:

* web services behind a 2FA login
* locally running software (Docker Desktop, for example)

but automating most would be better than automating none. (Somewhat in the same way that the link check we have covers 80-90% of URLs, but not all, because some sites block GitHub Actions runner IPs.)

This also sidesteps the personal choice of how you make your screenshots.

docs/publishing-repo.md
StevenMaude: Ah, right, now I got it; it's when the reader is copy-pasting!

> You can update your repo by replacing the relevant sections in your own README (typically just the first section) with the content from the template.

Maybe:

> The [current README template](https://github.com/opensafely/research-template/blob/main/README.md) has the updated version for new repos.
> To update your repository's README to match this new README:
> 1. Copy the content from the template repository's latest README (typically [just the first section](https://github.com/opensafely/research-template/blob/main/README.md?plain=1#L1-L3)).
> 2. Manually edit those changed sections into your research repository's README.
> 3. You will need to replace the two placeholder strings in the README's "View on OpenSAFELY" URL to point to your repository:
>
> * `${GITHUB_REPOSITORY_OWNER}` is the GitHub organisation your repo lives under, likely `opensafely`.
> * `${GITHUB_REPOSITORY_NAME}` is the name of your repo.

Something like that would have made it clearer for me. I'm not entirely sure that linking the lines directly is ideal, because they could change though. And there might be a more concise way of putting that.

StevenMaude: thoughts:

* That escaped URL is a little painful to edit manually. It's *almost*, but not quite, like an "update README/template" action would be useful.
* Likewise, it might almost be useful versioning the research template, so you can link directly to a version.

ghickman: > a little painful to edit manually

Understatement of the year 😂

We're not sure if we'll update readmes to save the researchers time yet, but I wanted the docs to help any researcher who happens to read them between now and when we make a decision.  There's every possibility that we generate a list of every readme in the org and some lucky soul gets to go fix them all!

Issue #886, opened 2022-07-21
Unpin black installation


StevenMaude:
This was pinned in #885; it should be unpinned when possible.

It depends on the version of click used in [snex](https://github.com/jwbargsten/snex): we need snex to upgrade to a version of click that is also compatible with the current versions of black.

StevenMaude:
This is also annoying in that MkDocs depends on click, so to successfully commit code that changes Python, you may need:

* a `.pre-commit` virtualenv
* a documentation virtualenv, just to run the MkDocs server
* the Data Builder virtualenv, to run Data Builder

StevenMaude:
One issue here is: there is `just check` in the `databuilder/` directory and this uses the `.venv` inside `databuilder`.

StevenMaude:
A better temporary workaround might be to fork snex and update the click requirement to allow us to install the same version of black. But this is still a bit of a mess.

StevenMaude:
The actual fix for the broken pre-commit was to specify the requirement of `click` in `.pre-commit-config.yaml` as in #900.

We still should unpin this requirement when possible, so leaving open.

StevenMaude:
I think the fix here is to move away from snex to the default support now added to Python Markdown; it removes the extra dependency: https://facelessuser.github.io/pymdown-extensions/extensions/snippets/#snippet-lines :smiley:

PR #1011, opened 2022-10-31
Remove unnecessary black config

ghickman:
By default (according to --help as of v22.8.0) black is already ignoring this list of directories:

/(\.direnv|\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|venv|\.svn|_build|buck-out|build|dist|__pypackages__)/

We were using --exclude which overwrote this list.  Black supports --extend-exclude which builds upon this list and also makes use of the repos .gitignore [1].  Since we we want both git and black to ignore the files for this repo can drop all black config here.

1: https://black.readthedocs.io/en/stable/usage_and_configuration/file_collection_and_discovery.html#gitignore

Refs: opensafely-core/repo-template#85

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,15 +1,3 @@
-[tool.black]
-exclude = '''
-(
-  /(
-      \.git         # exclude a few common directories
-    | \.direnv
-    | \.venv
-    | venv
-  )/
-)
-'''
-
 [tool.isort]
 profile = "black"
 lines_after_imports = 2


StevenMaude:
In principle, this is fine, Note that we're [currently pinned to a slightly older of Black](https://github.com/opensafely/documentation/issues/886) due to a conflicting requirement: I'll look at #886 and #1015 when there's a minute, and then we should be able to get up-to-date.

StevenMaude:
Just to explain here, because it took me a minute to understand: as mentioned in opensafely-core/repo-template#85, if we remove the `exclude` as this PR does, then the default behaviour for the current version of Black *is* to exclude the files listed in `.gitignore`.

StevenMaude (approved):

ghickman:
That's right.  Using `--exclude` tells black "exclude these paths and only these paths".  When we don't specify that it uses the default list, any paths in .gitignore and optionally anything we specify in `--extend-exclude`.

PR #1038, opened 2022-11-22
Clarify Docker installation

StevenMaude:
And also fix some trailing whitespace.

We previously recommended a Docker Desktop installation with Hyper-V.

This is presumably because:

* Docker Toolbox was not as polished
* WSL2 was still new, at the time of writing (late 2020)

Advantages to the changes here:

* we remove the mention of Docker Toolbox, which is now deprecated
* WSL2 works on Windows Home editions
* there's no confusion over our guidance saying "use Hyper-V" and Docker Desktop's installer saying "use WSL2"

### DIFF
--- a/docs/install-docker.md
+++ b/docs/install-docker.md
@@ -1,39 +1,59 @@
 
 ## Why Docker?
 
-Docker allows you to run identical software on all platforms. 
+Docker allows you to run identical software on all platforms.
 It creates "containers" that are guaranteed to be identical on any system that can run Docker.
 
-OpenSAFELY uses Docker to run your code in a reproducible, safe manner. 
+OpenSAFELY uses Docker to run your code in a reproducible, safe manner.
 This is most helpful for checking that you will be able to successfully run your code on the OpenSAFELY server on real data.
 If you only run your code locally using your own installation of R, say, then you won't know if the version of R (and the packages) installed on the server will run your code without errors or unexpected behaviours.
 See the [Testing Your Code section](actions-pipelines.md) for more details on how to test your code in practice.
 
-Unfortunately, Docker is happiest on Linux; on Windows and Mac OSX, installation can be a chore. 
+Unfortunately, Docker is happiest on Linux; on Windows and Mac OSX, installation can be a chore.
 These notes should help.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


StevenMaude:
The caveat here is that I haven't got a Windows install to test WSL2 with. I *assume* it works because it's the recommended default for Docker Desktop on Windows, and I don't believe we're doing anything particularly exotic in Windows.

One thing is that Windows 10 doesn't support GUI apps via WSL2 (Windows 11 does), but I don't think that is likely to affect us.

If it is, I can reword to still just suggesting Hyper-V.

docs/install-docker.md
```diff
@@ -1,40 +1,60 @@
...
-If so, [follow these instructions](https://github.com/docker/for-win/issues/785#issuecomment-344805180).
-It is best to install using the default settings.
+    Docker Desktop's recommended configuration is now the WSL2 backend.
+    We have not thoroughly tested OpenSAFELY with WSL2,
+    but there is no reason why WSL2 should be incompatible with OpenSAFELY's tooling.
 
-You may be asked to enable the Hyper-V and Containers features, which you should do.
-You can do this by [following these instructions](https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/quick-start/enable-hyper-v). 
+    You only need to choose either Hyper-V or WSL2.
+    It is fine to continue to use the Hyper-V backend.
```
iaindillingham, line 34: As a reader, this makes me nervous. The Docker people say "WSL2". The OpenSAFELY people say "WSL2, but we haven't tested it, and Hyper-V is fine".

Could we (as OpenSAFELY people) commit to WSL2, and keep a very close eye out for issues?

iaindillingham (approved):

docs/install-docker.md
StevenMaude: > Could we (as OpenSAFELY people) commit to WSL2, and keep a very close eye out for issues?

This is probably the right thing to do seeing as that's where Docker looks to be headed.

I'd be happier if a friendly Windows user could test WSL2 so that we know it does work. (I don't expect that it won't. But confirmation would be great :heavy_check_mark:)

I don't have a native Windows install to test with, unfortunately.

PR #1040, opened 2022-11-23
add link to blog

brianmackenna:
We have found that users get confused between pseudoBNF and dm+d.

Feedback has been that people have found this blog helpful https://github.com/opensafely/documentation/discussions/890

I think this section could benefit from a much longer rewrite but I propose that later as things are likely to change as a result of codelist rot.

### DIFF
--- a/docs/codelist-creation.md
+++ b/docs/codelist-creation.md
@@ -131,7 +131,7 @@ How contributions to codelists are acknowledged -- to be agreed.
 This is a short video showing how to build a pseudoBNF medication codelist. For OpenSAFELY studies you can convert this
 to NHS Dictionary of Medicines and Devices codelist which we briefly mention at the end of the video. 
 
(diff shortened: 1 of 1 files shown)


docs/codelist-creation.md
```diff
@@ -129,7 +129,11 @@ How contributions to codelists are acknowledged -- to be agreed.
 ## <a name="medvid"></a>Build a simple medication codelist
 
 This is a short video showing how to build a pseudoBNF medication codelist. For OpenSAFELY studies you can convert this
-to NHS Dictionary of Medicines and Devices codelist which we briefly mention at the end of the video. More on this soon!
+to NHS Dictionary of Medicines and Devices codelist which we briefly mention at the end of the video. 
+
+Very briefly, we recommend that you and agree medication codelists using the pseudoBNF coding system and convert this to
```
StevenMaude, line 134: @brianmackenna: This all looks fine, though is there a missing item after "and"?

HelenCEBM (approved):

PR #1054, opened 2022-12-08
Specify `site_url` in MkDocs configuration for correct sitemap

### LINKED ISSUE #1053: `sitemap.xml` not correctly generated for production site
Locally, it works:

```
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
    <url>
         <loc>http://127.0.0.1:8000/</loc>
         <lastmod>2022-12-08</lastmod>
         <changefreq>daily</changefreq>
    </url>
    <url>
         <loc>http://127.0.0.1:8000/access-policies/</loc>
         <lastmod>2022-12-08</lastmod>
         <changefreq>daily</changefreq>
    </url>
…
```

But on the [production site](https://docs.opensafely.org/sitemap.xml),

```
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
    <url>
         <loc>None</loc>
         <lastmod>2022-12-08</lastmod>
         <changefreq>daily</changefreq>
    </url>
    <url>
         <loc>None</loc>
         <lastmod>2022-12-08</lastmod>
         <changefreq>daily</changefreq>
    </url>
…
```

The fix is to specify `site_url: https://docs.opensafely.org` in `mkdocs.yml`.

However, I'm not sure we want this setting globally: strictly speaking, for local development and temporary Cloudflare PR, the site URL is different.

It is possible to [make this switchable depending on deployment context, as I previously mentioned](https://github.com/opensafely/documentation/pull/327#issuecomment-883324756); I hadn't realised that this issue broke the sitemap.

StevenMaude:
Fixes #1053.

To correctly generate the `sitemap.xml`, the [`site_url` must be specified in `mkdocs.yml`](https://github.com/mkdocs/mkdocs/issues/1783).

Having a valid sitemap can help search engines [index a site's content](https://developers.google.com/search/docs/crawling-indexing/sitemaps/overview).

What I've done here is to make `site_url` switchable depending on an environment variable, `MKDOCS_SITE_URL` and then simply set that on build. As this is only really required for deployment, I've just put this in the deployment workflow, instead of adding some wrapper or to do the build with the environment variable supplied.

This keeps the local development setup exactly as it was, with no value set for `site_url`.

I'm reasonably convinced that this is the right approach because:

* we can override the `site_url`
  * to set as an empty string if we ever want to build a [local file distribution](https://www.mkdocs.org/user-guide/deploying-your-docs/#local-files) (this may not be necessary with a missing value, but not tested)
  * when running locally with no URL set, the configuration is as it was previously (although I think with `mkdocs serve`, the site URL is ignored anyway!)
* we probably do want the canonical URL to be `https://docs.opensafely.org/some-path…` even for Cloudflare test deployments, technically, should a search engine somehow stumble upon those pages

### DIFF
--- a/.github/workflows/pages-deployment.yml
+++ b/.github/workflows/pages-deployment.yml
@@ -25,7 +25,7 @@ jobs:
           python3 -m pip install -r ./requirements.txt
 
       - name: Build site
-        run: mkdocs build
+        run: MKDOCS_SITE_URL="https://docs.opensafely.org" mkdocs build
 
       - name: Add a version file
         run: echo ${{ github.sha }} > site/version.html
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,4 +1,5 @@
 site_name: OpenSAFELY documentation
+site_url: !ENV MKDOCS_SITE_URL
 repo_url: https://github.com/opensafely/documentation
 nav:
   - Introduction: index.md


StevenMaude:
> I'm reasonably convinced that this is the right approach…

…but am happy to be persuaded otherwise, that, for instance, we should only set `site_url` for actual production deployments to the site (and not for other Cloudflare Pages previews that we get)

inglesp (approved):
Looks reasonable.

PR #1075, opened 2023-01-04
Corrected code for creating manual codelist

alschaffer:
Removed "column = " argument since it is not needed

### DIFF
--- a/docs/study-def-codelists.md
+++ b/docs/study-def-codelists.md
@@ -51,7 +51,7 @@ In some cases you may only want a variable to use a one or two codes, e.g you wa
 
 ```py
 weight_codes = codelist(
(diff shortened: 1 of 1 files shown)


docs/study-def-codelists.md
```diff
@@ -51,8 +51,8 @@ In some cases you may only want a variable to use a one or two codes, e.g you wa
 
 ```py
 weight_codes = codelist(
-    ["27113001", "162763007"], system="snomed", column="code"
+    ["27113001", "162763007"], system = "snomed"
```
rebkwok, line 54:
```suggestion
    ["27113001", "162763007"], system="snomed"
```

rebkwok (approved):
One suggested change for python formatting

PR #1090, opened 2023-01-12
Databuilder docs spike

rebkwok:
Spike to pull in databuilder docs from the databuilder repo.
Currently uses the `docs-spike` branch on databuilder.

This requires minimal config and pretty much just works from the parent docs repo (i.e. this one). It does require that we have an entry in the nav for the databuilder docs to live - currently I've put it under the "Using OpenSAFELY" section.


mkdocs_theme.yml
```diff
@@ -0,0 +1,33 @@
+site_name: OpenSAFELY documentation
```
rebkwok, line 1: This file isn't required for the main docs, but at the moment it's the only way I found to get the docs to run from databuilder and pull in the parent config

mkdocs.yml
```diff
@@ -132,6 +133,8 @@ plugins:
       custom_templates: templates
   - mkdocs-opensafely-databuilder
   - macros
+  - multirepo:
+      cleanup: false
```
rebkwok, line 137: for development only; in production we'd set this to true so it cleans up the imported files after building the site

PR #1094, opened 2023-01-12
Add a simple validation of Cloudflare redirects

StevenMaude:
We don't check these, and nor does Cloudflare indicate when they are broken.

So check them regularly.

We could also validate that they redirect to the correct URL, but checking the redirect code matches is possibly enough for now.

(Checking the final URL is possibly complicated if that URL itself redirects; not sure what Cloudflare's behaviour would be here, would it do an intermediary redirect or optimise by directly redirecting to the eventual final URL?)


StevenMaude:
Working in my personal fork:

A screenshot of check-redirects workflow passing in GitHub Actions.

PR #1110, opened 2023-01-27
Add basic introduction to `opensafely exec` cli command

### LINKED ISSUE #185: Proposal: Add info on making repos public
This may well be classed as something that should fall under people's existing git knowledge, but personally would find it super useful if there was a guide somewhere for the steps one would typically go through before making a repo public, and when we expect things to be made public in the OS workflow. 

For example, I know how to make a repo public, that this happens when I preprint and that I need to update the README, but I'm unsure about the role of releases and when we'd typically use those...

bloodearnest:
Fixes https://github.com/opensafely-core/opensafely-cli/issues/185

### DIFF
--- a/docs/opensafely-cli.md
+++ b/docs/opensafely-cli.md
@@ -117,16 +117,76 @@ See the [Codelist](codelist-intro.md) section for more information on codelists.
 
 
 To run your code on your machine, the `opensafely` tool uses the same Docker
-images that run in the secure server environments. These are updated
-periodically, for example when new libraries are installed. If you have error
-messages about missing libraries, your Docker images may need upgrading.
-To pull the most recent Docker images to your machine, run:
+images that run in the secure server environments. There is the
+`cohortextractor` image, for processing study definitions, and then the `r`,
+`stata-mp`, and `python` images, for running your analysis code. These last
+three provide a pre-built environment for their specific language, with
+a fixed set of pre-installed libraries.
+
+
+These are updated periodically, for example when new libraries are installed.
+If you have error messages about missing libraries, your Docker images may need
+upgrading.  To pull the most recent Docker images to your machine, run:
 
 ```bash
 opensafely pull
 ```
 
-### `jupyter` - running JupyterLab
+
+### `exec` - Interactive development
+
+Normal development of analysis code uses the pipeline defined in project.yaml to execute and test your code.
+
+However, data science languages are often used interactively to rapidly experiment and test code. The `opensafely exec` command provides a simple way to do this, using the Docker images provided by OpenSAFELY. This can help ensure that your code works correctly in OpenSAFELY as you develop it, rather than accidentally relying on tools and libraries installed on your own machine.
+
+Running `opensafely exec IMAGE COMMAND` does the following:
+ - runs an instance of the appropriate docker IMAGE (`r`, `python`, `stata-mp`)
+ - shares the files in your current directory with the instance
+ - executes COMMAND (or the default command for the image if you don't supply one)
+
+This allows you develop and and test code as if it was a regular interactive session.
+
+For example, to run an interactive Stata session:
+
+```bash
+opensafely exec stata-mp
+```
+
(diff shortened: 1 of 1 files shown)


docs/opensafely-cli.md
```diff
@@ -117,15 +117,75 @@ See the [Codelist](codelist-intro.md) section for more information on codelists.
...
+opensafely exec stata-mp
+```
+
+This will run the Stata packaged in the `stata-mp` docker image, and you can manually test your Stata code. (the opensafely tool knows how to fetch and apply the OpenSAFELY Stata licence).
+
+
+Likewise, for R:
+
+```bash
+opensafely exec r R
```
bloodearnest, line 160: Note: this won't work until the new R image is out, but that's hopefully soon.

docs/opensafely-cli.md
```diff
@@ -117,15 +117,75 @@ See the [Codelist](codelist-intro.md) section for more information on codelists.
...
 
 To run your code on your machine, the `opensafely` tool uses the same Docker
-images that run in the secure server environments. These are updated
-periodically, for example when new libraries are installed. If you have error
-messages about missing libraries, your Docker images may need upgrading.
-To pull the most recent Docker images to your machine, run:
+images that run in the secure server environments. There is the
+`cohortextractor` image, for processing study definitions, and then `r`,
+`stata-mp`, and `python` images, for running your analysis code. These last
+three provide a pre-built environment for their specific tool, with a set of libraries.
```
sebbacon, line 123: I found the terminology "tool" a bit confusing here - I'd find "language" more natural.

docs/opensafely-cli.md
```diff
@@ -117,15 +117,75 @@ See the [Codelist](codelist-intro.md) section for more information on codelists.
...
+These are updated periodically, for example when new libraries are installed.
+If you have error messages about missing libraries, your Docker images may need
+upgrading.  To pull the most recent Docker images to your machine, run:
 
 ```bash
 opensafely pull
 ```
 
+
+### Developing code with the OpenSAFELY images.
```
sebbacon, line 135: This heading feels like it implies it's describing the usual way of developing any code. Is that what we're recommending, now? I had thought it was more a debugging tactic, but may have this wrong.

If it's more about debugging alternatives, how about "Using the OpenSAFELY images from the command line" or "Using the OpenSAFELY images outside the pipeline"? And adding an intro sentence along the lines "usually you'll develop your code by running it throuhg the opensafely pipeline [link], but sometimes it's useful to....<blah>"?

docs/opensafely-cli.md
```diff
@@ -117,15 +117,75 @@ See the [Codelist](codelist-intro.md) section for more information on codelists.
...
+
+The R and Stata tools are often used interactively to develop and test analysis code, and python users often use `ipython` or `jupyterlab` to develop and test code.
+
+The `opensafely exec` command provides a simple way to do this, by running interactive tools using the Docker images provided by OpenSAFELY. This can help ensure that your code works correctly in OpenSAFELY as you develop it, rather than accidentally relying on tools and libraries installed on your own machine.
```
sebbacon, line 139: Per my previous point - perhaps we want to be encouraging a pipeline-first approach, in which case it would be worth adding something like "The pipeline automatically ensures the languages and tools used on your own machine match those in OpenSAFELY, but if you're debugging problems outside the pipeline...."

milanwiedemann (approved):
Thanks, I think this is great.

docs/opensafely-cli.md
milanwiedemann: this is where we could link to a future R demonstration

docs/opensafely-cli.md
bloodearnest: "Interactive development with the OpenSAFELY environments"?

I think "debugging" is not the right language, as could indicate its very technical rather than generally useful

bloodearnest:
@sebbacon are you happy with the changes?

sebbacon (approved):
Yes, it's great.

Noticed some typos this time round

PR #1126, opened 2023-02-09
update allowed file types

LFISHER7:
This expands on the documentation for allowed file types, with some considerations for each output type. In particular, it highlights when html files are appropriate for release and how they should be formatted. They are difficult to review, and often could be produced as individual tables/figures, so these suggestions raise the bar for them being requested.

The updated text references some instructions on producing stripped html files which don't yet exist, so I'll leave this in draft until they do.


docs/releasing-files.md
```diff
@@ -165,16 +165,37 @@ If you are likely to release data multiple times, e.g. for initial discussion wi
 
 ## 2. Requesting release of outputs from the server
 
-Having applied disclosure controls to the aggregated study data, you are ready to request their release. **Only specific members of the OpenSAFELY team trained in output checking have permissions to release the data**. When you are ready to request a release of your aggregated results, and only the minimum outputs necessary please [complete this form](documents/OpenSAFELY_Output_Review_Form_07-04-22.docx) and email us at **<datarelease@opensafely.org>**. For each output wishing to be released you will need to provide a clear contextual description including:
+Having applied disclosure controls to the aggregated study data, you are ready to request their release. **Only specific members of the OpenSAFELY team trained in output checking have permissions to release the data**. When you are ready to request a release of your aggregated results, and only the minimum outputs necessary please [complete this form](documents/OpenSAFELY_Output_Review_Form_07-04-22.docx). For each output wishing to be released you will need to provide a clear contextual description including:
```
Jongmassey, line 168: what's the rationale for removing this email address?

Jongmassey:
Here's a thought....

Given that there's a simple pattern to finding script blocks in HTML (i.e. from `<script>` to `<\script>` ... ish) - could we make this part of osrelease? I'm thinking of a workflow like a lot of the pre-commit checking used by the tech team - i.e. you try to release the file and it warns about, then corrects the error and prompts you to review:

1. reviewer tries to release files including html files
2. osrelease checks html files for `<script>` tags
  a. if found, prints warning about script tags
  b. copies file to staging folder, having removed the script blocks
  c. reviewer checks the files in the staging folder
  d. reviewer osrelease's the files in the staging folder
3. continues with release

Jongmassey:
or a "html clean" reusable action would be nice

docs/releasing-files.md
```diff
@@ -165,16 +165,37 @@ If you are likely to release data multiple times, e.g. for initial discussion wi
...
+    * If you are producing a lot of figures, consider combining them into panel plots or requesting release of the underlying data so that you can produce them outside of the secure server.
+* **Other**
+    * `txt` files can be released, but you should consider whether the output can be produced as a table, which is easier to review.
+    * `json` files can be released, but as with tables, make sure that the attributes are easily understandable for reviewers. If the output can be represented as a table, you should consider converting it.
+    * `html` files can be released if you are producing a report style output, but please note the points below:
+        * `html` files are harder to review than other output types, so should be reserved for reports which require both contextual text and embedded outputs. Most commonly, this will be a report to be hosted on [OpenSAFELY Reports](https://reports.opensafely.org/) (more information [here](https://docs.opensafely.org/reports/intro/)). If you can produce your report locally, using individually released files, you should.
+        * Hide any code blocks from the report.
```
Jongmassey, line 190: is "Hide" the right term here? "Remove" or is "Hide" the terminology used by  rmd/jupyter  etc?

docs/releasing-files.md
LFISHER7: I've moved this into a section below on submitting the form to separate out filling out the form and what it should include, from the submission process

docs/releasing-files.md
LFISHER7: I think you're right "hide" isn't quite right here, but we don't want people to actually "remove" code blocks. I'll change this to something along the lines of "Make sure the code blocks aren't rendered in the final html if you don't need them. You can do this by...."

LFISHER7:
I like the idea of a reusable action to do the cleaning. That has the advantage over making changes via osrelease that what a user requests for release is the same as what is actually released.

I've previously made a script for converting a jupyter notebook [here](https://github.com/opensafely/notebook-action/blob/initial-release/analysis/notebook_action.py), but didn't progress it as we thought better documentation for nbconvert would be sufficient. Perhaps an action that supports "clean" conversion of jupyter and r mrakdown files would be useful

Jongmassey:
The more I think about it, the simple answer is just to disallow all HTML outputs.

LFISHER7:
I think we're probably heading that way. The only real requirement we have for them is to support the reports site.

I'd quite like to get this merged, unless you have any more comments?

docs/releasing-files.md
```diff
@@ -167,16 +167,37 @@ There is also a disclosure control section in our [Q&A forum](https://github.com
...
+    * `html` files can be released if you are producing a report style output, but please note the points below:
+        * `html` files are harder to review than other output types, so should be reserved for reports which require both contextual text and embedded outputs. Most commonly, this will be a report to be hosted on [OpenSAFELY Reports](https://reports.opensafely.org/) (more information [here](https://docs.opensafely.org/reports/intro/)). If you can produce your report locally, using individually released files, you should.
+        * Make sure that any code blocks are not rendered in the rendered report if they are not needed. You can find examples showing how to do this for Jupyter notebooks and R markdown files [here](https://docs.opensafely.org/reports/intro/#producing-reports). 
+        * Each individual output within the report should be requested for release separately, with the contextual information outlined above.
+        * `html` files should be stripped of any embedded javascript and styling. This is obfuscated when viewing a report via a web browser, but makes review of the raw file very difficult. You can find instructions on how to do that [here](TODO)
```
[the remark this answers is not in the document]
Jongmassey: I don't think we should be committing todos to the docs, raise an issue that it is to be done instead?

Jongmassey (changes requested):

docs/releasing-files.md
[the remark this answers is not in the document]
LFISHER7: Thanks for spotting, I've replaced this with the relevant link

Jongmassey (approved):

PR #1127, opened 2023-02-09
add r markdown and jupyter instructions

LFISHER7:
Adds instructions on how to produce reports in OpenSAFELY using R markdown and Jupyter notebooks. This accompanies recommendations around producing html files in #1126.

This should address the following feedback on what would be useful:

- how to run jupyter in opensafely
- Basic markdown and point them to a cheatsheet
- Adding figures - with example python
- Adding tables - with example python (pandas)
- Running jupyter notebooks as an action

### DIFF
--- a/docs/reports/intro.md
+++ b/docs/reports/intro.md
@@ -58,6 +58,8 @@ text = display(
     Markdown("Some text which can be Markdown formatted.")
     )
 ```
+
+Details on formatting markdown cells can be found [here](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet).
 ### Converting Jupyter notebooks to html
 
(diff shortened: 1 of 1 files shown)


LisaHopcroft:
I've pushed a mini R markdown example. Not sure if the proper syntax highlighting is working though, I'll investigate.

LFISHER7:
Thanks, that's useful. Does `kable` add any funky formatting? These examples would ideally render html that doesn't have any javascript or styling added that makes output checking difficult

LisaHopcroft:
No, `kable` is very basic. But we could replace `kable` with `default` if you want the very plainest version.

LisaHopcroft:
I can't find a way to force the markdown to render in a way that is more specific for R markdown, so I think that's the best we can get.

LisaHopcroft:
Looks good to me!

LisaHopcroft (approved):

PR #1135, opened 2023-02-15
Use `just` to install requirements in workflow

StevenMaude:
Forgot this in #1134.

### DIFF
--- a/.github/workflows/pages-deployment.yml
+++ b/.github/workflows/pages-deployment.yml
@@ -22,13 +22,8 @@ jobs:
           install-just: true
           python-version: '3.9'
 
-      - name: Upgrade pip
-        run: |
-          python3 -m pip install --upgrade pip
-
-      - name: Install dependencies
-        run: |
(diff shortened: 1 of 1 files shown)


iaindillingham (approved):

.github/workflows/pages-deployment.yml
```diff
@@ -22,13 +22,8 @@ jobs:
...
 
-      - name: Upgrade pip
-        run: |
-          python3 -m pip install --upgrade pip
-
-      - name: Install dependencies
-        run: |
-          python3 -m pip install -r ./requirements.txt
+      - name: Install Python requirements
+        run: just devenv
```
ghickman, line 26: It looks like `just build` already does this?

StevenMaude: yes, you're right and I'm half-asleep! So we can just remove the lines, which is better.

iaindillingham: I'm greater than half asleep. Sorry for not spotting this. Thanks, George!

PR #1139, opened 2023-02-21
Rename Cloudflare API token secret

StevenMaude:
To both prevent name collisions, and make it more obvious what the token is for.

### DIFF
--- a/.github/workflows/pages-deployment.yml
+++ b/.github/workflows/pages-deployment.yml
@@ -41,7 +41,7 @@ jobs:
         uses: cloudflare/pages-action@c2ad89679f0dee1ff4bcf9b8bbe1cf52d176cb29  # v1.3.0
         with:
           accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
-          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          apiToken: ${{ secrets.CLOUDFLARE_DIRECT_UPLOAD_API_TOKEN }}
           directory: "site"
           gitHubToken: ${{ secrets.GITHUB_TOKEN }}
           projectName: "opensafely-docs"


.github/workflows/pages-deployment.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         uses: cloudflare/pages-action@c2ad89679f0dee1ff4bcf9b8bbe1cf52d176cb29  # v1.3.0
         with:
           accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
```
bloodearnest, line 40: I changed CLOUDFLARE_ACCOUNT_ID to CLOUDFLARE_API_TOKEN.

Did I missunderstand the brief?

StevenMaude: I think `CLOUDFLARE_ACCOUNT_ID` should stay as is.

StevenMaude: Or, for the most clarity, what's required here are organisation secrets in both `opensafely` (for this repository) and `opensafely-core` (for `databuilder`) with the names:

* `CLOUDFLARE_ACCOUNT_ID` — as it was in this repository, our Cloudflare account's ID
* `CLOUDFLARE_DIRECT_UPLOAD_API_TOKEN` is a new name but with the same value as `CLOUDFLARE_API_TOKEN` in this repository

bloodearnest: Ok, I think we just need to rename CLOUD_FLARE_API_TOKEN to CLOUDFLARE_ACCOUNT_ID, as I misread the original suggestions

bloodearnest: I've renamed `s/CLOUDFLARE_API_TOKEN/CLOUDFLARE_ACCOUNT_ID/` in both orgs and in bitwarden.

bloodearnest: I've delete CLOUDFLARE_API_TOKEN from both org secrets, so it should just be this repos version that needs removing.

StevenMaude: @bloodearnest: thanks!

evansd (approved):

PR #1147, opened 2023-02-27
Adding guidance for adding a paper to OpenSAFELY.org

LisaHopcroft:
Internal users will have access to this information via the Team Manual. We have included the information here so that external users can prepare a `.md` file that their co-pilot can then incorporate into the website.


ccunningham101 (approved):
Could clarify the navigation tab, but also probably fine as-is

PR #1178, opened 2023-03-29
update form

LFISHER7:
This replaces the review request form with an updated version that includes instructions on requesting the release of log files


amirmehrkar (approved):
Correct

PR #1185, opened 2023-03-31
Build site to pull in databuilder repo before checking links

rebkwok:
Upgrading this repo to 3.11 broke the link checking in the ehrql repo. Since that link-checking actually did the checking on a checkout of this repo, it makes sense to just put the ehrql link checking here too.

- build site first so the link checker also checks ehrql links
- increase frequency of workflow run to nightly (ehrql changes regularly, we want to check links more frequently than once a week)
- Add the notify slack job for failures


rebkwok:
This will need access to a `SLACK_BOT_TOKEN` secret

evansd (approved):

PR #1189, opened 2023-04-03
add midpoint 6 rounding

LindaNab:
Explanation of midpoint 6 rounding and example code.


LFISHER7 (approved):
Thanks for adding this @LindaNab. I've made a few small changes to presentation, but I think the content is good. I'd like to share this with someone, so I'm going to go ahead and merge it but we can update it with any feedback we get

PR #1234, opened 2023-06-06
Reinstate scheduled docs check every Tuesday

StevenMaude:
We tried using a Slack reminder, but you can't control which specific weekdays a Slack reminder appears, if the reminders aren't weekly.

(You can schedule reminders like "on the 15th day of every month", or "every Tuesday" for example, but not "every fourth Tuesday".)

This is fairly inflexible and ended up being posted on a Sunday. Which no-one spotted.

Instead, go back to running the job automatically, but reduce the frequency to once per week on a given weekday, which seems reasonable. Previously, the job ran automatically once per day, when removed, which would be irritating if no-one is dealing with any persistent issue immediately.

Running automatically also removes the need for someone to first read the reminder, then manually run the workflow and *then* have the error appear in Slack.


.github/workflows/check_links.yml
```diff
@@ -4,6 +4,8 @@ name: 'Check links'
 # yamllint disable-line rule:truthy
 on:
   workflow_dispatch:
+  schedule:
+    - cron: '37 11 * * 2'
```
iaindillingham, line 8: 11:37's a bit... precise? But no matter 🙂

StevenMaude: I tend to pick random times, so as to reduce the chance of clashing with other workflows. The default people would go for is probably 0 minutes.

(That said, GitHub's scheduled times are not precise anyway: a trigger might be set for 11:00 and run at some random time a few minutes after.)

PR #1236, opened 2023-06-06
Covid-19 DPN DPIA 2023 security tier diagram

amirmehrkar:


iaindillingham (commented):
I think you probably want to replace [the current image](https://docs.opensafely.org/security-levels/#diagram) with this image, no? If so, then it should be called `OpenSAFELY-security-levels.svg`. If you're a bit stuck, then let me know.

iaindillingham (approved):

PR #1238, opened 2023-06-06
Explain how the link check process works

StevenMaude:
How to find broken URLs, and how to fix them.

This is based on my suggestion that someone else sees this process in opensafely-core/ehrql#1244.

***

:warning: However, it may be that this content is more suitable for the Bennett team manual, because we also use this same process for other sites we maintain.

If that's so, it's OK to suggest to move it there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,3 +3,69 @@
 This is the public documentation for using the [OpenSAFELY platform](https://www.opensafely.org/).
 
 It provides information on how to get set up with and use the platform.
+
+## Link checking
+
+To help keep the content up-to-date,
+this repository has a scheduled weekly link check run with lychee.
+
+### Finding the inaccessible URLs
+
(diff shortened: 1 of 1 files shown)


StevenMaude:
Probably needs some formatting tidying.

Also, not sure if this is in `README.md`, or it should go in `DEPLOY.md` (and make that `DEVELOPERS.md`?)

StevenMaude:
:warning: It may be that this content is more suitable for the Bennett team manual, because we also use this process for other sites.

If that's so, it's OK to suggest to move it there.

iaindillingham (approved):
This is great; really useful. I think it belongs here, in the README, because:

* It's a public demonstration of how seriously we take our documentation (it belongs in this repo)
* It's not just for the tech team; anyone can help fix broken URLs (it belongs in this file)

PR #1246, opened 2023-06-13
Updating output checking processes

LFISHER7:
Adds updates to our stance on rounding and some instruction on releasing intermediate results

resolves https://github.com/ebmdatalab/service-analytics-team/issues/131
resolves https://github.com/ebmdatalab/service-analytics-team/issues/142

### DIFF
--- a/docs/releasing-files.md
+++ b/docs/releasing-files.md
@@ -113,7 +113,7 @@ If you are unsure about anything, please email us: [disclosurecontrol@opensafely
 The redaction of any counts <=5 reduces the risk of primary disclosure, but it does not remove the risk of secondary disclosure. As there are many active projects using the OpenSAFELY platform and datasets, there is a risk that your results might be combined with other results to reveal information about individuals or groups. To reduce this risk, we ask that you **round any counts to the nearest 5**. This includes rounding counts underlying any figures requested for release. In most cases this is unlikely to have a significant impact on your results. If it does, please highlight why it is important to release non-rounded counts when making an output review request.
 
 !!! note
-    Rounding does not remove the need to redact counts <=5. You should first redact any low counts and then apply rounding. If applying both techniques, the recommended approach is to first redact counts <=7 and then round to the nearest 5 (if you redact counts <=5 and then round to the nearest 5, any counts of 5 remaining must originally have been either 6 or 7; this approach provides the same protection for all counts)
+    Rounding does not remove the need to redact counts <=5. You should first redact counts <=7 and then round to the nearest 5 (if you redact counts <=5 and then round to the nearest 5, any counts of 5 remaining must originally have been either 6 or 7; this approach provides the same protection for all counts)
 
 Below is an example of a table before (top) and after (bottom) rounding has been applied. **Note that column and row totals are the sum of the rounded counts.**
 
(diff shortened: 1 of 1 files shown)


docs/releasing-files.md
```diff
@@ -136,7 +136,7 @@ Below is an example of a table before (top) and after (bottom) rounding has been
...
+A rate consists of a numerator and denominator, which are generally both counts. **In OpenSAFELY, any rate calculated from counts below the redaction threshold (5 or less) should also be a redacted.** In addition, we recommend rounding because redaction alone is vulnerable to differencing. When future calculations rely on rates not being mapped to a non-numerical like `[REDACTED]` and/or a distinction between a rate of zero and a non-zero rate is desirable, we recommend rounding the numerator and denominator to 'midpoint 6'. In short, rounding to 'midpoint 6' allows differentiating between zero and non-zero rates, by not breaking our suppression rules and without introducing bias.
```
HelenCEBM, line 139:
```suggestion
A rate consists of a numerator and denominator, which are generally both counts. **In OpenSAFELY, any rate calculated from counts below the redaction threshold (5 or less) should also be redacted.** In addition, we recommend rounding because redaction alone is vulnerable to differencing. When future calculations rely on rates not being mapped to a non-numerical like `[REDACTED]` and/or a distinction between a rate of zero and a non-zero rate is desirable, we recommend rounding the numerator and denominator to 'midpoint 6'. In short, rounding to 'midpoint 6' allows differentiating between zero and non-zero rates, by not breaking our suppression rules and without introducing bias.
```

docs/releasing-files.md
```diff
@@ -226,6 +226,19 @@ For each output wishing to be released you will need to provide a clear contextu
 
 Each section in the review request form should normally describe a single file, but where necessary for similar files, these can be grouped together and wildcards can be used for the file path (e.g. `release/hospitalisation_rate_by_*.csv`). **If you use a wildcard, please indicate how many files this captures**.
 
+### Release of intermediate data
+
+In general, releases should be for final results from your project (see the note above). However, on some occassions it is appropriate to release intermediate data. Below are some suggestions for when this is appropriate:
+
+* You think you may need to make minor edits to final outputs such as changing figure labels. Release of the intermediate data allows you to make these changes locally.
+* A large number of outputs are produced from a single intemediate output. Release of the intermediate data underlying the figures (which needs to be checked whether it is released or not) avoids the need to check the downstream outputs.
+*  The intemediate data doesn't contain person-level data, but is used for running a model that would produce multiple outputs.
```
HelenCEBM, line 235:
```suggestion
*  The intermediate data doesn't contain person-level data, but is used for running a model that would produce multiple outputs.
```

docs/releasing-files.md
```diff
@@ -226,6 +226,19 @@ For each output wishing to be released you will need to provide a clear contextu
 
 Each section in the review request form should normally describe a single file, but where necessary for similar files, these can be grouped together and wildcards can be used for the file path (e.g. `release/hospitalisation_rate_by_*.csv`). **If you use a wildcard, please indicate how many files this captures**.
 
+### Release of intermediate data
+
+In general, releases should be for final results from your project (see the note above). However, on some occassions it is appropriate to release intermediate data. Below are some suggestions for when this is appropriate:
+
+* You think you may need to make minor edits to final outputs such as changing figure labels. Release of the intermediate data allows you to make these changes locally.
+* A large number of outputs are produced from a single intemediate output. Release of the intermediate data underlying the figures (which needs to be checked whether it is released or not) avoids the need to check the downstream outputs.
```
HelenCEBM, line 234:
```suggestion
* A large number of outputs are produced from a single intermediate output. Release of the intermediate data underlying the figures (which needs to be checked whether it is released or not) avoids the need to check the downstream outputs.
```

docs/releasing-files.md
```diff
@@ -226,6 +226,19 @@ For each output wishing to be released you will need to provide a clear contextu
...
+*  The intemediate data doesn't contain person-level data, but is used for running a model that would produce multiple outputs.
+
+If requesting release of intermediate data there are a few considerations:
+
+*  We recommend that you continue to develop downstream analysis actions within the OpenSAFELY pipeline, even if they are not inteneded to be run on the server against. This helps maintain reproducibility.
```
HelenCEBM, line 239:
```suggestion
*  We recommend that you continue to develop downstream analysis actions within the OpenSAFELY pipeline, even if they are not intended to be run on the server. This helps maintain reproducibility.
```

HelenCEBM:
I wonder if the intermediate data part could be a little clearer, e.g. re. the first point on changing chart labels, I would normally release the data behind a chart anyway (e.g for supplementary data) so would consider that final data rather than intermediate... I'm not sure if this is the first mention of running code outside rather than inside the server so not sure if some people might get a bit confused by that?

LFISHER7:
I agree by our definition of final outputs on this page (as being outputs for your publication), the underlying data could be considered a final output. But it's not been common for people to request release of this. Are you able to suggest how this could be clearer? 

By the time people are releasing outputs, they should be familiar with the concept of running locally but i've added in a section with some considerations for when you run locally on released outputs.

docs/releasing-files.md
```diff
@@ -292,6 +308,14 @@ These outputs can be shared with project collaborators and published in line wit
     - transcribing (e.g., to paper or email)
     - using screen sharing software or any recording device/software
 
+### Running further analyses on released outputs
+
+If you have had [intermediate data released](#release-of-intermediate-data) and you wish to run further analyses on them, such as reformatting figures, there are a few things to consider.
+
+1. You should include the code for these steps in your GitHub repo.
+2. You **should not** commit any of the released outputs to your GitHub repo. Make sure to include them in the `.gitignore` file.
```
HelenCEBM, line 316: what about the final processed charts/tables created as a result, should they be excluded from the repo as well?

LFISHER7: I think they have to be, for now, given our approaches to making repos public and the template readme content. Ultimately, we'll need to be able to upload results to the jobs site

Issue #1289, opened 2023-07-26
Check for links to `docs.opensafely.org` in PRs


StevenMaude:
Spotted in #1290. I've fixed these there, but there's nothing to stop anyone reintroducing them in future.

Contributors may not realise that the preferable approach is linking internally to the documentation, rather than the published version of that site, so that the URLs are to the correct version of the documentation wherever it is built and run. It does not help that the Markdown URL syntax is easy to get wrong when linking to files. 

We could probably use Vale or similar — see #643 — to check for this and suggest to someone making changes that they probably want a relative Markdown link.

PR #1290, opened 2023-07-26
Tidy links a little

StevenMaude:
* Ditch links to `https://docs.opensafely.org`; these should be internal links within the Markdown.
* Remove use of "here" in link text as it is unhelpful, [particularly for users using screen readers](https://usability.yale.edu/web-accessibility/articles/links#disabilities):
  > While screen readers can read a full page to a user, screen reader users may prefer to instead listen to a list of links. In that case, a screen reader may only read the link text and not the surrounding text.
* Fix some small whitespace issues either manually in surrounding sentences that have been edited, or in files that have been edited as found by pre-commit.


StevenMaude:
While it's fine to manually clear these up now and again, we should try and nudge contributors to make the correct changes. I've opened #1289 and #1291 for this purpose.

StevenMaude:
![Link check running in GitHub Actions and passing for commit 79de8c98bdd0ad2ea83992f3d087dce4612801b5.]https://github.com/opensafely/documentation/assets/3818079/fecacbbf-3b04-4bc0-801d-769064f49481)

tomodwyer (approved):

PR #1299, opened 2023-08-01
Add basic Vale configuration

StevenMaude:
Fixes #1289.
Fixes #1291.
Relates to #519 and #643.

This adds a minimal Vale configuration as a trial.

With this change, adding rules can help us be more consistent across our documentation. These are not enforced anywhere, but can be checked optionally by one or more of:

* running Vale locally against the documentation source via `just lint-docs`
* installing Vale into editor/IDE for suggestions while editing
* running the `check_docs` workflow, included in this PR that allows you to list the suggestions for a branch

The rules may still need further refinement especially as they are regex-based, but they are not enforced anywhere.

Running Vale with these rules on the documentation source results in 0 suggestions, since these were fixed in #1304.

:warning: This PR *does not*:

* add some automated way of installing Vale locally; individuals must do that themselves.
* check the ehrQL documentation; that will need some thinking about integration, particularly because:
  * ehrQL documentation is pulled in separately from the ehrQL repository
  * a sizeable chunk of the ehrQL documentation is autogenerated from Python docstrings

PR #1304, opened 2023-08-03
Make various small improvements found by Vale

StevenMaude:
Based on #1299.

Most of the changes are found from the checks proposed in #1299. There are some other small fixes, either caught by pre-commit (trailing whitespace in edited files) or by eye in the surrounding text.


StevenMaude:
This doesn't completely clear all the Vale suggestions. That's because the style rules need some further refinement: particularly around false positives of OpenSAFELY URLs being lowercase "opensafely", which is actually fine.

PR #1306, opened 2023-08-07
Tidy 'publication of outputs on jobserver' sections

HelenCEBM:
Mention that copilot should chase up on outputs being turned from draft to public


LFISHER7 (approved):

HelenCEBM:
Note similar text also appears on https://docs.opensafely.org/jobs-site/#publishing-outputs so perhaps we should link to that instead of repeating here.

LFISHER7 (approved):

PR #1332, opened 2023-09-14
Move a couple of sections to top level in sidebar

inglesp:
This slightly increases the depth of content on display, without negatively impacting navigability.


inglesp:
Before:

image

After:

image

PR #1383, opened 2023-11-16
Update Actions section to reference ehrQL

inglesp:
Addresses part of https://github.com/opensafely-core/ehrql/issues/1676

### DIFF
--- a/docs/actions-pipelines.md
+++ b/docs/actions-pipelines.md
@@ -5,7 +5,7 @@ This section covers how to develop, run, and test your code to ensure it will wo
 
 ## Project pipelines
 
-The [cohortextractor](actions-cohortextractor.md) section describes how to make an action which generate dummy datasets based on the instructions [defined in your `study_definition.py` script](study-def.md).
+The [ehrQL](/ehrql/) documentation describes how to make an action which generate dummy datasets based on the instructions defined in your `dataset_definition.py` script.
 These dummy datasets are the basis for developing the analysis code that will eventually be passed to the server to run on real datasets.
 The code can be written and run on your local machine using whatever development set up you prefer (e.g., developing R in RStudio).
 However, it's important to ensure that this code will run successfully in OpenSAFELY's secure environment too, using the specific language and package versions that are installed there. To do this, you should use the project pipeline.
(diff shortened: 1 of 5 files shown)


mkdocs.yml
```diff
@@ -42,7 +42,6 @@ nav:
       - Actions:
           - Overview: actions-intro.md
           - The project pipeline: actions-pipelines.md
-          - The cohortextractor action: actions-cohortextractor.md
```
StevenMaude, line 45: question: Is this page to be removed? Or is the intent just to remove it from the navigation alone (which is fine)?

docs/actions-pipelines.md
```diff
@@ -5,7 +5,7 @@ This section covers how to develop, run, and test your code to ensure it will wo
 
 ## Project pipelines
 
-The [cohortextractor](actions-cohortextractor.md) section describes how to make an action which generate dummy datasets based on the instructions [defined in your `study_definition.py` script](study-def.md).
+The [ehrQL](/ehrql/) documentation describes how to make an action which generate dummy datasets based on the instructions defined in your `dataset_definition.py` script.
```
StevenMaude, line 8: suggestion: Maybe linking to the dummy data page for ehrQL dataset definitions would be better?

```suggestion
The [ehrQL](/ehrql/how-to/dummy-data.md) documentation describes how to make an action which generate dummy datasets based on the instructions defined in your `dataset_definition.py` script.
```

(I've not checked the correctness of that relative link, but the page is better than linking to a more general part of the ehrQL documentation, I think.)

StevenMaude, line 8: thought: There's also a slight quirk here where, previously, `study_definition.py` encompassed measures or what we now call datasets, whereas `dataset_definition.py` explicitly refers to datasets (and presumably someone would use `measure_definition.py` instead for measures).

I don't think probably matters too much though.

StevenMaude (approved):

mkdocs.yml
inglesp: I'm going to salvage some possibly-useful cohortextractor stuff for some legacy docs.

StevenMaude: Great :+1:

docs/actions-pipelines.md
inglesp: I don't think these docs are tight enough to make this worth worrying about.

PR #1398, opened 2023-11-29
Replace Gitpod with GitHub Codespaces

### LINKED ISSUE #1173: Remove references to Gitpod
**Update**: we now want to remove Gitpod entirely.

---

@brianmackenna pointed out that the [getting started guide](https://docs.opensafely.org/getting-started/#options-for-running-opensafely) in the docs implies that gitpod is the preferred method of running OpenSAFELY. It currently is not - at least not for committed users.  If we continue to support gitpod we may want to consider:

- stating clearly that approved OpenSAFELY users (as opposed to interested parties just having a look at the tool) should aim to use a local install if possible
- Switching the order of either the introduction to the options, or the tabs (or both) so that the local install options are shown first

StevenMaude:
Fixes #1173.

Our Gitpod configuration for the [research template stopped working and has now been removed](https://github.com/opensafely/research-template/pull/109). Instead, we currently have a configuration that works with GitHub Codespaces.

The Getting Started guide could not be followed by users trying to use OpenSAFELY via Gitpod. This PR updates the Getting Started guide to use Codespaces.

We originally added Gitpod when Codespaces wasn't available for all. We now favour GitHub Codespaces because Codespaces doesn't require:

* an extra user account to be setup, on Gitpod (everything is on GitHub)
* linking of that extra Gitpod account to the user's GitHub account, to enable repository access (Codespaces already enables this with no additional steps)
* auditing and approval of another web service in addition to GitHub by IT administrators who manage hardware of OpenSAFELY users, although access to Codespaces might still have to be requested
* that open virtual machines are managed via a site other than GitHub

All this simplifies the process for users, and our explanations of that process.

We could consolidate some of the other [GitHub Codespaces page](https://steve-replace-gitpod-with-gi.opensafely-docs.pages.dev/github-codespaces/) with this one, as they now overlap, but that can be a future PR.


StevenMaude:
I think a subsequent improvement would be to consolidate the GitHub Codespaces page with the Getting Started guide. There is some overlap there. We could merge in the sections that refer to creating a demonstration research repository to the Getting Started guide, or remove them entirely.

But this PR at least drops the mention of Gitpod, so that could be a subsequent PR either shortly or when we revisit the Getting Started guide.

StevenMaude:
There's a tiny bit of work here to decide what to do with the other screenshots that were taken, and probably weren't needed. But the content changes could be reviewed now.

iaindillingham (approved):

StevenMaude:
I'll leave the screenshots as they are. If we find cropping some of them is better, we could do that in future.

Issue #1410, opened 2023-12-12
Update notes on Python Docker image


StevenMaude:
The Python Docker image has now been split into v1 and v2 images: the v1 being the previously existing image, and v2 a newer image with upgraded Python and packages. 

This week's link check flagged this up (see below for broken URLs); the documentation should be updated to reflect this. This also indicates the affected pages.

***

```
Errors per input
Errors in docs/requesting-libraries.md

    [404] https://github.com/opensafely-core/python-docker/blob/main/requirements.in | Failed: Network error: Not Found

Errors in docs/actions-reusable.md

    [404] https://github.com/opensafely-core/python-docker/blob/main/requirements.txt | Failed: Network error: Not Found

Errors in docs/actions-scripts.md

    [404] https://github.com/opensafely-core/python-docker/blob/main/requirements.txt | Failed: Network error: Not Found
```

StevenMaude:
My — possibly limited — understanding of the specific changes here are something like:

* there's a v1 image which is the old image for compatibility with old studies where users don't want to upgrade (this is still `latest`, again for backwards compatibility, so that existing studies using `latest` don't unintentionally use the v2 image)
* the v1 and v2 images now have specific versions that indicate the date and time (YYMMDDHHSS) of release (e.g. `v1.231211121504)
* the v1 image won't be updated (except for essential security fixes, maybe?); if you want newer software then use v2
* the v2 image has upgraded Python and Python packages
* breaking changes in future will result in a new major version of the image
* v1 is based on Ubuntu 20.04; v2 is based on Ubuntu 22.04
* the requirements for both images are now in v1 and v2 directories in the repository (and update the URLs)

That may not necessarily be all information that end users need to know.

I'll tag @bloodearnest in case Simon has anything to add here too.

bloodearnest:
In fact, the link should probably be https://github.com/opensafely-core/python-docker/blob/main/v1/packages.md

StevenMaude:
Thanks!

I'll update the broken links to move, and leave this issue open for us to update the advice to users.

PR #1411, opened 2023-12-14
Fix links to Python Docker image

StevenMaude:
See #1410. These were moved in the recent revisions to how the Docker image is built.


docs/actions-scripts.md
```diff
@@ -34,7 +34,7 @@ Scripted actions can read and write output files that are saved in the workspace
 
 Outputs labelled `highly_sensitive` will not be visible to researchers. This is a [deliberate design feature of OpenSAFELY](https://www.opensafely.org/about/), intended to reduce the risk of disclosure of sensitive information. Outputs should **always** be classed as `highly_sensitive` if they are:
 
-- Pseudonymised patient-level outputs derived from queries run against Level 1 and 2 data, i.e., a specific study dataset generated by a [dataset_definition](https://docs.opensafely.org/ehrql/).
+- Pseudonymised patient-level outputs derived from queries run against Level 1 and 2 data, i.e., a specific study dataset generated by a [dataset definition](ehrql/index.md#a-dataset-definition).
```
inglesp, line 37: Oops, it was probably me who used an absolute URL.

But I don't think we should be linking to halfway down the ehrQL index page.  If somebody knows what a dataset definition is, they won't need to follow the link; if they don't, it'd be better to take them to the ehrQL index page.  But this is a judgement call, so feel free to disregard.

inglesp (approved):

PR #1465, opened 2024-03-21
remove cohortextractor content from APC page

alexwalkerepi:
I came across these examples, and initially assumed there'd be more `cohortextractor` content in the /data-sources/ section, but I think they might have been removed in a previous purge.

I suspect that these still remain because there are useful snippets of info about the data within it, such as reference tables for how some variables are coded, and these are [referenced in the ehrQL docs](https://docs.opensafely.org/ehrql/reference/schemas/tpp/#apcs).

I have tried to remove the `cohortextractor` parts while retaining the information that is not also in the ehrQL docs.

I think in future we should consolidate the info in the ehrQL schema reference, the data-sources section, and the [database reports here](https://reports.opensafely.org/) into one place.


evansd (approved):
This looks good to me, but I realised after I started reviewing that this is a #team-rex repo and they should probably have the final say here.

PR #1473, opened 2024-04-03
Ignore link to Project Update form

StevenMaude:


iaindillingham (approved):

StevenMaude:
I think that this form is intentionally left as non-viewable to the general public, but have asked Liam to confirm this, before I'll merge this.

StevenMaude:
Closing because the form is now public, and the link check passes without this change.

PR #1478, opened 2024-04-11
Restructure Getting Started section

StevenMaude:
See opensafely-core/codespaces-initiative#40.

We are going to follow the [Diátaxis framework](https://www.diataxis.fr) in this Getting Started section, as we have done with the ehrQL documentation. This is a prelude to editing the Getting Started guide (opensafely-core/codespaces-initiative#39).

The tutorial has been placed in a table of contents mirroring the ehrQL structure. There is only one page — the previous Getting Started guide — but we may split this up in the near future into individual pages that describe each logical step.

Specifically:

* The long "Getting Started" guide is moved into a "Getting Started/Tutorial" section. This is the only page right now. In future PRs, we could split this up into multiple pages, one per logical step, following the structure of the ehrQL tutorial.
* The "Getting Started" guide is now referred to as a tutorial, and links to it updated.
* The installation guides are moved around in the table of contents. *The associated documentation source Markdown files are not moved in the repository yet: this will change their URLs. Since these pages might be moved, removed or merged in future, I've left them for now until they are further reviewed.*
* The installation "Overview" page is now merged into the "Getting Started/How-to guides" and removed; there is a redirect added for this removed page.
* There is some minimal added information to give context to the new structure on the "landing" pages for each section.

PR #1485, opened 2024-04-17
Remove non-Codespaces guidance from Getting Started tutorial

### LINKED ISSUE #1396: Consider only using Codespaces for Getting Started guide
There are enough new concepts and steps required, that facilitating a local installation when someone is first trying out OpenSAFELY is perhaps an extra step too far.

The branching also makes reading the document clunky because there are two paths for most of the steps.

By using only Codespaces, this would:

* make it clearer for the reader to know what they are doing, and give them only one set of instructions
* make it easier for the user to get to success in this guide — editing a study repository, pushing it to GitHub and seeing the tests pass — because Codespaces is less likely to give problems than installing software
* give the user a better idea of the steps needed for success, before they start trying to install the required software
* reduce the total length of the Getting Started source document
* allow us to add more screenshots along the way; it's not easy to do this right now because some of the instructions are shared for users running OpenSAFELY in a browser and locally 

The reader could be directed to another tutorial or explanation of the required software, and guidance for installation.

StevenMaude:
Fixes #1396.

Partly addresses opensafely-core/codespaces-initiative#30.

### DIFF
--- a/docs/getting-started/explanation/index.md
+++ b/docs/getting-started/explanation/index.md
@@ -1,1 +1,3 @@
 This section contains background knowledge on the software used to run OpenSAFELY.
+
+* [Options for running OpenSAFELY](options-for-running-opensafely/index.md)
--- a/docs/getting-started/explanation/index.md
+++ b/docs/getting-started/explanation/index.md
@@ -1,0 +1,1 @@
+This section contains background knowledge on the software used to run OpenSAFELY.
--- a/docs/getting-started/explanation/index.md
+++ b/docs/getting-started/explanation/index.md
@@ -1,3 +1,4 @@
 This section contains background knowledge on the software used to run OpenSAFELY.
 
 * [Options for running OpenSAFELY](options-for-running-opensafely/index.md)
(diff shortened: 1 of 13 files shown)


StevenMaude:
Depends on #1478.

StevenMaude:
In this PR, I am pulling out the non-Codespaces content from the Getting Started tutorial, and either:

* merging it with the duplicated content, if it is a duplicate of existing information
* storing it in the most relevant place, if it is not a duplicate of existing information

That's not to say that we should keep everything that's moved here, but it:

* cleans up the existing tutorial to streamline it to relevant information only
* highlights what has been moved out, giving us a way to decide to keep or delete

StevenMaude:
Another improvement would be to split the tutorial into pages that represent each step, like the ehrQL tutorial. So instead of numbered sections in a single page, there would be one page for each section, and maybe an introductory page.

I'm not sure that's worth doing right now, as we may remove most or all of what we have in the near future.

StevenMaude:
I also think that any newly added pages should probably be added to the table of contents too. As it's not yet decided what should be included, I haven't done so.

iaindillingham (approved):

PR #1502, opened 2024-05-01
Remove references to adding a paper

tomodwyer:
As this is no longer required.


StevenMaude (approved):
:partying_face:

brianmackenna:
I'm not sure this is the correct etiquette - please forgive me!

I think we do need something for the docs on this - we have a lot of external contributors to our website..............even if Bennett people are adding the things on their behalf - it would be good to have guidance on how to prepare the information to reduce burden on Bennett people!

PR #1508, opened 2024-05-03
Fix dummy data links

benbc:
There are broken links to the dummy data documentation page [here](https://docs.opensafely.org/actions-pipelines/#project-pipelines) (start of first para) and [here](https://docs.opensafely.org/workflow/) (item 3).

The fixed pages are [here](https://benbc-fix-dummy-data-links.opensafely-docs.pages.dev/actions-pipelines/#project-pipelines) and [here](https://benbc-fix-dummy-data-links.opensafely-docs.pages.dev/workflow/).

### DIFF
--- a/docs/actions-pipelines.md
+++ b/docs/actions-pipelines.md
@@ -5,7 +5,7 @@ This section covers how to develop, run, and test your code to ensure it will wo
 
 ## Project pipelines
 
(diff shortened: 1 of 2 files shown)


evansd:
There's something I don't fully understand going on with links in our documentation. The fix you have here (which I'm quite sure _is_ a fix) is to remove the `.md` extension. But the fix which Steve applied to the links here is exactly the reverse:
 * https://github.com/opensafely-core/ehrql/pull/1813/files

I think we should probably get to the bottom of this one way or the other.

evansd (approved):

PR #1523, opened 2024-05-19
Add "How to use released outputs in GitHub Codespaces" page

iaindillingham:
Closes opensafely-core/codespaces-initiative#59

### DIFF
--- a/docs/getting-started/how-to/use-released-outputs-in-github-codespaces/index.md
+++ b/docs/getting-started/how-to/use-released-outputs-in-github-codespaces/index.md
@@ -1,0 +1,39 @@
+Whilst it's possible to execute a project pipeline entirely inside a secure environment,
+in practice it's often more convenient to execute the final stage outside.
+This is because the final stage frequently involves carefully crafting figures and tables for publication,
+making many small adjustments that would otherwise entail multiple round-trips to the OpenSAFELY Jobs site.
+
+Executing the final stage of a project pipeline outside a secure environment is only possible when the outputs from the previous stage have been released to the OpenSAFELY Jobs site.
+[Released outputs](../../../releasing-files-intro.md) have been subject to statistical disclosure control and have been reviewed by two trained OpenSAFELY output checkers.
+
+To upload released outputs to a Codespace, using VS Code:
+
+* Download the released outputs from the OpenSAFELY Jobs site to your computer.
+* Right-click on the `output` folder in the Explorer, which is in the Primary Side Bar.
+  The "[User Interface](https://code.visualstudio.com/docs/getstarted/userinterface)" page in VS Code's documentation locates the Explorer and the Primary Side Bar.
+* Click "Upload...".
+* Select the released outputs to upload.
+
+To download figures and tables from a Codespace, using VS Code:
+
+* Locate the figures and tables in the `output` folder in the Explorer, which is in the Primary Side Bar.
+  The "[User Interface](https://code.visualstudio.com/docs/getstarted/userinterface)" page in VS Code's documentation locates the Explorer and the Primary Side Bar.
+* Right-click on each figure or table.
+* Click "Download...".
+
+Uploading released outputs to a Codespace,
+and downloading figures and tables from a Codespace,
+are examples of **sharing results**.
+The "[Datasets Used](https://www.opensafely.org/policies-for-researchers/#datasets-used)" section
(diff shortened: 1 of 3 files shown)


docs/getting-started/how-to/use-released-outputs-in-github-codespaces/index.md
```diff
@@ -0,0 +1,30 @@
...
+in practice it's often more convenient to execute the final stage outside.
+This is because the final stage frequently involves carefully crafting figures and tables for publication,
+making many small adjustments that would otherwise entail multiple round-trips to the OpenSAFELY Jobs site.
+
+Executing the final stage of a project pipeline outside a secure environment is only possible when the outputs from the previous stage have been released to the OpenSAFELY Jobs site.
+[Released outputs](../../../releasing-files-intro.md) have been subject to statistical disclosure control and have been reviewed by two trained OpenSAFELY output checkers.
+
+To copy released outputs to a Codespace,
+first download them from the OpenSAFELY Jobs site to your computer,
+then drag-and-drop them from your computer to the Codespace.
```
Jongmassey, line 11: I think this needs a bit of expanding (and probably an image) to say to drop to the left sidebar file browser. When I test this myself - dragging a file from my desktop to the editor area opens the file in the editor (even if it's a file there's no handler for, e.g. pdfs just appear as garbled text), rather than putting the file onto the codespace's filesystem.

docs/getting-started/how-to/use-released-outputs-in-github-codespaces/index.md
```diff
@@ -0,0 +1,30 @@
...
+Copying released outputs to a Codespace,
+and copying figures and tables from a Codespace,
+are examples of **sharing results**.
+The "[Datasets Used](https://www.opensafely.org/policies-for-researchers/#datasets-used)" section
+of the "Policies for Researchers" page on the OpenSAFELY website contains more information about sharing results.
+
+!!! warning
+    Never commit released outputs, figures, or tables to your project's repository.
+    Doing so would contravene the "Policies for Researchers" page on the OpenSAFELY website.
+
```
Jongmassey, line 24: This feels like it would be very easy to do by accident. Should we recommend/ship a released outputs folder with appropriate gitignore and tell people to drag and drop there?

Jongmassey (changes requested):

docs/getting-started/how-to/use-released-outputs-in-github-codespaces/index.md
iaindillingham: Okay, thank you. I was hoping to avoid a screenshot but I don't think that's going to be possible 😕

lucyb: If you're providing screenshots (and I think it makes sense in this case), it might be easier to suggest people use the right-click menu and select the "upload" option, while also mentioning that drag and drop is an option.

lucyb (approved):
This looks good. I'm approving with the expectation that Jon's comments will be addressed.

docs/getting-started/how-to/use-released-outputs-in-github-codespaces/index.md
iaindillingham: I've updated the text, removing references to drag-and-drop and adding references to the "Upload..." and "Download..." menu items. I haven't included screenshots, because I think the "[User Interface](https://code.visualstudio.com/docs/getstarted/userinterface)" page in VS Code's documentation -- which I've linked to -- is thorough, clear, and updated by someone else 🙂 Do you have strong feelings about screenshots?

docs/getting-started/how-to/use-released-outputs-in-github-codespaces/index.md
Jongmassey: That sounds like a good approach 👍

docs/getting-started/how-to/use-released-outputs-in-github-codespaces/index.md
Jongmassey: I think uploading is safer than drag and dropping.  Re: screenshots I'm generally in the _picture painting a thousand words_ camp but appreciate they're unresponsive to future UI changes! Text seems fine now, we can consider screenshots if folks seem to struggle.

Jongmassey (approved):

PR #1535, opened 2024-05-31
Fix broken links in output request form

LFISHER7:


milanwiedemann (approved):
Not sure which links didn't before but I checked all the links in the updated .docx and they all work 🎉

LFISHER7:
Thanks!

PR #1561, opened 2024-07-03
Update notes on link checking

StevenMaude:
To help clarify the process for others who might need to follow this.


StevenMaude:
This is with a move to making this a tech support rota activity.

It may be too much detail, but I feel like specifying everything and then paring it back (either in this PR, or in future) is possibly the right approach.

iaindillingham (approved):
LGTM 🚀 Haven't looked at a rendered version. I expect the Cloudflare issue is transient.

PR #1566, opened 2024-07-10
Set Permissions-Policy to disable Topics

iaindillingham:
We set the Permissions-Policy header to disable Topics, which replace Google's [Federated Learning of
Cohorts](https://privacysandbox.com/proposals/floc/) (FLoC). For more information, see: ebmdatalab/sysadmin#354.

### DIFF
--- a/docs/_headers
+++ b/docs/_headers
@@ -1,0 +1,2 @@
+/*
+  Permissions-Policy: browsing-topics=()


iaindillingham:
Tested in Chrome 126.0.6478.127 by following the link to the branch preview and checking that `topics = await document.browsingTopics()` raises a `DOMException` as described by "[Document: browsingTopics() method](https://developer.mozilla.org/en-US/docs/Web/API/Document/browsingTopics)".

tomodwyer (approved):

PR #1572, opened 2024-07-17
Update `jupyter:latest` to `python:v2`

### LINKED ISSUE #1571: Replace mentions of `jupyter` image with `python:v2`
Related to #1568: we have some mentions of `opensafely jupyter` in the documentation still.

The OpenSAFELY [jupyter-docker image repository](https://github.com/opensafely-core/jupyter-docker) was archived in January 2024, so we should update the documentation and instead suggest using the python:v2 image.

Edit: see below. We do have a `run: jupyter` that should be fixed up though.

StevenMaude:
Fixes #1571.

### DIFF
--- a/docs/reports/intro.md
+++ b/docs/reports/intro.md
@@ -74,7 +74,7 @@ This removes any code blocks in the notebook from the rendered html; to keep the
 To run this within an OpenSAFELY action, you can use the following run command:
 
 ```
-run: jupyter:latest jupyter nbconvert /workspace/analysis/notebook.ipynb --execute --to html --output-dir=/workspace/output --no-input --ExecutePreprocessor.timeout=86400
+run: python:v2 jupyter nbconvert /workspace/analysis/notebook.ipynb --execute --to html --output-dir=/workspace/output --no-input --ExecutePreprocessor.timeout=86400
 ```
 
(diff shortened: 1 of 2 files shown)


docs/reports/intro.md
```diff
@@ -74,7 +74,7 @@ This removes any code blocks in the notebook from the rendered html; to keep the
 To run this within an OpenSAFELY action, you can use the following run command:
 
 ```
-run: jupyter:latest jupyter nbconvert /workspace/analysis/notebook.ipynb --execute --to html --output-dir=/workspace/output --no-input --ExecutePreprocessor.timeout=86400
+run: python:v2 jupyter nbconvert /workspace/analysis/notebook.ipynb --execute --to html --output-dir=/workspace/output --no-input --ExecutePreprocessor.timeout=86400
 ```
 
 Refer to an [example of this, implemented as an OpenSAFELY action](https://github.com/opensafely/mechanical-valve-anticoag/blob/1f158504ba5a74470b11c8d73311fb2859d67cb7/project.yaml#L53-L63).
```
StevenMaude, line 80: thought: Arguably, we should change this link to another example too, because it uses `jupyter:latest`.

There aren't any examples that use `python:v2`.

There is [at least one that uses `python:latest`](https://github.com/opensafely/tpp-database-schema/blob/bd0b21c453d249f55674eab7969aabd0dd59a412/project.yaml#L29-L52), although the actual example is simpler (notebook creates a single HTML output) compared with the example linked here (notebook creates multiple kinds of outputs).

We could address this here, but I also opened #1573 to record this case and others in the documentation.

iaindillingham (approved):

PR #1587, opened 2024-08-02
Update T1OO text to reflect new reality

evansd:
See preview at:
https://evansd-t1oo-update.opensafely-docs.pages.dev/type-one-opt-outs/#how-is-permission-to-access-type-1-opt-out-data-enforced


rebkwok (approved):
Lgtm, but probably wants an ops check too

PR #1605, opened 2024-08-22
Auto-append the glossary to every page

rebkwok:
Use the `auto_append` option for snippets so we don't need to include the glossary on every page.

This also means that the main docs glossary will be auto-appended to the imported Airlock docs (along with Airlock's own glossary).


StevenMaude:
Great; thank you!

StevenMaude (approved):

PR #1619, opened 2024-09-11
Fix links

rebkwok:
1) Fix some actually broken links and anchors. 
2) All digital.nhs.uk links are returning 403s now in the [links check](https://github.com/opensafely/documentation/actions/runs/10791686722), despite working before.  This seems to be a persistent failure (at least, it's been consistent for the last 24 hrs or so) I've verified that all the links are accessible in a browser, so have just added them to the ignore list.

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -19,6 +19,7 @@ https?://www\.encepp\.eu(?:/.*)?$
 https?://www\.tandfonline\.com(?:/.*)?$
 https?://www\.icnarc\.org(?:/.*)?$
 https?://doi\.org(?:/.*)?$
+https?://digital.nhs.uk(?:/.*$)?$
 
 # Twitter used to work but is being very flaky.
 # This exclusion should be reviewed in future to see if the issue persists.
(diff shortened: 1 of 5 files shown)


StevenMaude:
> All digital.nhs.uk links are returning 403s

Yep, I spotted this for other site checks we have too, and have skipped this check in the team manual too. I think it's fair to say this is deliberate blocking of bots.

.lycheeignore
```diff
@@ -27,6 +27,10 @@ https?://twitter\.com(?:/.*$)?$
 # An example OpenSAFELY project that is not intended to exist.
 https?://github\.com/opensafely/example-research-repository(?:/.*)?$
 
+# digital.nhs.uk links used to work but are now failing with a 403
+# Network error: Forbidden
+https?://digital.nhs.uk(?:/.*$)?$
```
StevenMaude, line 32: nit: I'd perhaps move this into the section above:

> `# Block either the GitHub Actions runner or the lychee client.`

rather than list it as a special case.

In my experience of updating `.lycheeignore`, this behaviour happens with quite a lot of websites. The checks initially pass, but later get blocked entirely.

(Whether sites are specifically targeting blocking of lychee or GitHub Actions IP address ranges, or just all bots as there are probably more aggressively scraping bots for AI content these days, I don't know.)

StevenMaude (approved):

StevenMaude:
Thanks for fixing these up :sparkles:

Issue #1626, opened 2024-09-19
Improvements to the getting started tutorial


iaindillingham:
Couple of suggestions from @Mary-Anya[^1] for the getting started tutorial:

* [Create a GitHub repository](https://docs.opensafely.org/getting-started/tutorial/create-a-github-repository/) at the bottom, "Create Repository from template" is now 'Create Repository'. The text and the screenshot should be updated.

* [Run the project pipeline](https://docs.opensafely.org/getting-started/tutorial/run-the-project-pipeline/) > Run the `project.yaml` file: add/repeat "Open project.yaml in the editor. This file will be near the end of the list of files and folders" - it's in the next section, but this is the first time you actually use the file.

[^1]: https://bennettoxford.slack.com/archives/C069SADHP1Q/p1726649680207919

lucyb:
I have volunteered @eli-miriam to pick this up, as they plan on making some related wording changes to the Getting Started tutorial based on their recent experience of going through it.

Issue #1623, opened 2024-09-18
Confusing prompt for additional permissions during the getting started tutorial


iaindillingham:
The getting started tutorial asks the user to [start a Codespace](https://docs.opensafely.org/getting-started/tutorial/create-a-github-codespace/) in a repo that was created from the research template. Having started a Codespace, and assuming the user's browser doesn't hand over to VS Code, the user is then prompted to authorise (or not) additional permissions. Whilst we have documented the prompt on the "[How to troubleshoot common issues in GitHub Codespaces](https://docs.opensafely.org/getting-started/how-to/troubleshoot-common-codespaces-issues/)" page, it's unlikely that the user will have read this page before starting the tutorial. Consequently, the prompt is confusing.

The prompt asks the user to authorise read permissions for [opensafely/server-instructions](https://github.com/opensafely/server-instructions), which is a private repository. The user is prompted even when they're _not_ a member of opensafely, as @Mary-Anya discovered. I'm surprised by this -- surely if the user isn't a member of opensafely, then they shoulnd't know about a private repository associated with opensafely.

We could...

* mention the prompt when the user is asked to start a Codespace
* start the Codespace with default permissions. (This is possible with [`gh codespace create`](https://cli.github.com/manual/gh_codespace_create).)

iaindillingham:
Martina asked about the prompt on Slack[^1] today. Thanks for answering her question, @Jongmassey.

[^1]: https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1726838657270489

Jongmassey:
It also might be possible to negate the need for this cross-repository permission by making using of a pre-existing organisation secret to set the env var required to make Stata play nice

PR #1633, opened 2024-09-27
Improvements to the Getting Started guide

eli-miriam:
Addresses #1623 and #1626, along with some other minor grammar, clarity and readability changes.

With regards the list in `run-the-project-pipeline`: I couldn't get it to display properly without resorting to HTML, but we use that elsewhere in this tutorial so I'm hoping that's an acceptable fix.

I also opted to remove the leading `$` from terminal commands, both for consistency with the ehrQL tutorial, and because it recently tripped up Linda (who is understandably unfamiliar with it as a convention).


iaindillingham (approved):
Nice! Thank you for the PR.

Issue #1573, opened 2024-07-17
Update links to example research code where that code is old and unmaintained


StevenMaude:
In #1572, it was noted that there is at least one link to research code that doesn't do things exactly as might be the OpenSAFELY best practice today (it uses `run: jupyter:latest` and not `run: python:v2`).

Other links to code repositories are currently in:

- [ ] `docs/federation.md`
- [ ] `docs/protocol.md`
- [ ] `docs/study-def-flowcharts.md`
- [ ] `docs/data-sources/apc.md`
- [ ] `docs/data-sources/ecds.md`
- [ ] `docs/data-sources/systmone.md`
- [ ] `docs/data-sources`cpns.md`
- [ ] `docs/study-def-tricks.md`

If these are outdated, we should replace them with more current examples.

lucyb:
I've [spoken to Brian](https://bennettoxford.slack.com/archives/C02HJTL065A/p1721814370984009) and someone in the research team will be able to update this code.

wjchulme:
**no action needed**
* The links from the data source pages are fine.
* The `study-def-flowcharts.md` and `study-def-tricks.md` are now under legacy so can stay as is.

**non-urgent action needed**
* The walkthrough in `docs/federation.md` should be replaced with an example in ehrQL, when such an example exists. The text currently acknowledges that the example is outdated as it uses study definitions, so think that enough until we can fully replace.
* The `docs/protocol.md` page links to protocols in repos, not code, so this is fine. Though we may want to update to more recent examples at some stage. 
* The `report/intro.md` page includes one link to [code that is out-of-date](https://github.com/opensafely/mechanical-valve-anticoag/blob/1f158504ba5a74470b11c8d73311fb2859d67cb7/project.yaml#L53-L63)

alschaffer:
Don't know if this fits here, but some references to modifying study def variables in cohortextractor here: 
_"Making changes to the study definition variables"_
https://docs.opensafely.org/updating-the-docs/

I'm not even sure what that paragraph is about so unsure if it needs updating, or deleting (I suspect the latter though).

StevenMaude:
> I'm not even sure what that paragraph is about so unsure if it needs updating, or deleting (I suspect the latter though).

It's not really helpful now as it doesn't apply to ehrQL, so I deleted it in #1642.

rebkwok:
[...]

PR #1639, opened 2024-10-08
Update `project.yaml` examples to pipeline v4 if the example only has ehrQL actions

alarthast:
- Part of [upgrading pipeline to version 4](https://github.com/orgs/opensafely-core/projects/15/views/1?pane=issue&itemId=81337702&issue=opensafely-core%7Cpipeline%7C224)
- Version 4 of pipeline removes support for cohort extractor and the `project.yaml` file should not have an `expectations` section.
- `case-control-studies.md` has examples that use cohort extractor actions so nothing has been changed there. Is there a reason why this page wasn't updated to use ehrQL?

### DIFF
--- a/docs/actions-pipelines.md
+++ b/docs/actions-pipelines.md
@@ -28,11 +28,7 @@ It is written using a configuration format called [YAML](https://yaml.org/), whi
 A simple example of a `project.yaml` is as follows:
 
 ```yaml
-version: "3.0"
-
-# Ignore this `expectations` block. It is required but not used, and will be removed in future versions.
-expectations:
-  population_size: 1000
+version: "4.0"
 
 actions:
   generate_dataset:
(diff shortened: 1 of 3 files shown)


docs/actions-pipelines.md
```diff
@@ -60,7 +56,7 @@ The `run_model` action will run a Stata script called `model.do` based on the `d
 It will output two moderately sensitive files `cox-model.txt` and `survival-plot.png`, which can be checked and released if appropriate.
 
 
-Every `project.yaml` requires a `version`, `expectations`, and `actions` section.
+Every `project.yaml` requires a `version` and `actions` section. An `expectations` section was also required for version 3.0 of the pipeline framework (example [here](case-control-studies.md)), but is not used in the latest version.
```
rebkwok, line 59:
```suggestion
Every `project.yaml` requires a `version` and an `actions` section.
```

I don't think we should reference cohort-extractor at all here; we've removed all references to it from everywhere else (except a historical reference to it in the type 1 opt outs description, and this one example). Although that case controls example uses a cohort-extractor action (and therefore a 3.0 project.yaml), there are no docs on any of the actual code there. Anyone looking at that page for the first time has no reference for anything in it, so I think we should just leave it out of here too.

Related: that case-controls-study example should be updated, but that's not for this PR

rebkwok (approved):
One suggestion to remove reference to cohort-extractor/v3.0

docs/actions-pipelines.md
rebkwok: I've added a comment to #1573

alarthast: Suggestion committed, given that the case control studies page is being worked on.

PR #1705, opened 2025-01-17
Incorporate v2 of the r image in the docs

remlapmot:
Fixes https://github.com/opensafely-core/r-docker/issues/184

PR #1711, opened 2025-01-27
Fix broken links for r:v1 package data

bloodearnest:
In preparing for v2: we changed this, and the links broke.

Also worth noting that those are md links now, not csv, like pythons
package version info, and are better formatter, with links.


remlapmot:
I did these and some other small fixes in #1705 , but can close that if you like

bloodearnest:
Once this lands, you should be able to rebase yours on it?

remlapmot:
Can do.

PR #1718, opened 2025-02-06
Check internal links only

### LINKED ISSUE #1709: Link check internal links only
We've had the new link checking process in place for a little while now and it appears to fail often. The link checker was recently configured to ignore all 50x errors and 429 errors (PR https://github.com/opensafely/documentation/pull/1671). However, transient errors when checking the links still happen frequently. This causes false positive alerts and makes it more likely people will ignore them, and generally just creates more work for people.

Therefore, I propose that we only check the validity of links for the docs.opensafely.org domain itself. Potentially, we could expand this to other sites that we control (since we're able to control any errors we receive from our own sites).

StevenMaude:
Fixes #1709.

### DIFF
--- a/.github/workflows/check_links.yml
+++ b/.github/workflows/check_links.yml
@@ -42,7 +42,6 @@ jobs:
       - name: Check links
         uses: lycheeverse/lychee-action@f613c4a64e50d792e0b31ec34bbcbba12263c6a6  # v2.3.0
         with:
-          args: "--exclude-all-private --include-verbatim --max-concurrency 24 --require-https --verbose --no-progress
+          args: "--exclude-all-private --include-verbatim --max-concurrency 24 --require-https --verbose --no-progress --offline
           --accept '100..=103,200..=299,429,500..=511'
           --timeout 60 './docs/**/*.md' './docs/**/*.html' './imported_docs/**/*.md' './imported_docs/**/*.html' --exclude-path './docs/ehrql/includes/generated_docs'"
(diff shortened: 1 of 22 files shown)


StevenMaude:
This passes the link check, though I noticed there are several warnings that lychee flags (that are also flagged by MkDocs) where links aren't exactly in the correct format (though may work). For example, missing the trailing `.md`. (This is an existing problem though.)

StevenMaude:
This gets rid of most of our MkDocs warnings, except the spam from the `mkdocstrings` package. If that ever gets a configuration option to silence those warnings, or we ever remove the cohort-extractor documentation that uses `mkdocstrings`, we could do a regular check of `mkdocs build --strict` to flag problems, without necessarily enforcing passing on PR merge (#764).

mikerkelly (approved):
Thanks Steve, looking good.

Nit feedback: I think the first couple commit messages could have been slightly more verbose and explicit to help anyone reading them in future understand more readily why a commit has been made (and without referring out to GitHub)

- "To focus on our own content." -> "We decided to check only internal links as checking external links produced too many false positives and was too much work."?
- "Remove use of GitHub token for link check" add second line "As we now pass the `--offline` flag, we don't need to contact GitHub"? I also think this would have been clearer if if the first two commits were merged.

StevenMaude:
I've rebased to combine the first two commits.

tomodwyer (approved):

PR #1724, opened 2025-02-12
Move footer buttons to content area

### LINKED ISSUE #1513: Previous and Next links
A user[^1] reports that it's easy not to notice the Previous and Next links on documentation pages. These links are especially important when a collection of documentation pages form a logical sequence, such as the "[Tutorial](https://docs.opensafely.org/getting-started/tutorial/)" collection in the "[Getting started](https://docs.opensafely.org/getting-started/)" section. (This collection was created by #1503.)

We should test this collection with a couple of users to establish whether the issue is commonplace. If it is, then we should consider a couple of candidate solutions before choosing one and implementing it.

[^1]: https://bennettoxford.slack.com/archives/C069SADHP1Q/p1715083970948039

tomodwyer:
Closes #1513

URLs starting with the path names listed in the [`buttonPaths` array of `extra.js`](https://github.com/opensafely/documentation/pull/1724/files#diff-471feeb498f79e7b563705cc384c369e28a28bde530c4734bc604fbced7a7a11R58) have buttons moved from the footer, to the main content area.

### DIFF
--- a/docs/js/extra.js
+++ b/docs/js/extra.js
@@ -37,3 +37,30 @@ function patchCopyCodeButtons() {
 document.addEventListener("DOMContentLoaded", () => {
   patchCopyCodeButtons();
 });
+
+/**
+ * Move the existing footer buttons to the main content section
+ * @returns {void}
+ */
+function moveNavButtons() {
+  /** @type {HTMLDivElement} */
+  const contentArea = document.querySelector(`[data-md-component="content"]`);
+  /** @type {HTMLElement} */
+  const footerNav = document.querySelector(`[aria-label="Footer"]`);
+  /** @type {HTMLElement} */
+  const footClone = footerNav.cloneNode(true);
+  footClone.classList.add("footer-nav-buttons");
+  contentArea.appendChild(footClone);
+  footerNav.setAttribute("hidden", "true");
+}
+
+/** @type {string[]} */
+const buttonPaths = ["/getting-started/tutorial/", "/ehrql/tutorial/"];
+/** @type {string} */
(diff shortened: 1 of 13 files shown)


docs/js/extra.js
```diff
@@ -37,3 +37,30 @@ function patchCopyCodeButtons() {
 document.addEventListener("DOMContentLoaded", () => {
   patchCopyCodeButtons();
 });
+
+/**
+ * Move the existing footer buttons to the main content section
```
mikerkelly, line 42:
```suggestion
 * Move the existing footer buttons to the main content section to make them more visible.
 * Material MKDocs doesn’t seem to have any options to do this natively.
```

mikerkelly (approved):
This all looks good, thanks.

I only wondered a little bit why it's scoped to the two tutorials. I can see that's probably where it's _most_ useful as they follow a linear progression and when you get to the bottom you want it to be intuitive which UI element to interact with next, but there are other pages in section groups that have the footer-style Previous/Next. Could/should we apply this across everything? 

[Original Slack thread](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1716470942697499?thread_ts=1715083970.948039&cid=C069SADHP1Q) mentions a page that I think has moved [here](https://docs.opensafely.org/outputs/releasing-overview/) for example.

tomodwyer:
@mikerkelly those two sections were the ones I was sure needed this functionality. I will add a commit now to add that extra section too.

My hope was that as we identified sections, or changed URLs, people would update the list of URLs.

Jongmassey (approved):

PR #1744, opened 2025-03-18
Update docs about ehrQL output formats

### LINKED ISSUE #1691: Out of date ehrql info
The section at https://docs.opensafely.org/actions-scripts/#large-highly_sensitive-output-files has some out of date information

- In the template, ehrql now outputs arrow files by default, not csv.gz
- Code examples still uses .feather exenstions, where as we recommend .arrow
- python code uses pandas, to_feather and read_feather, may need updating
- stata code doesn't give examples of loading arrow at all

rebkwok:
Arrow is now the preferred output format.
Adds some extra info about reading arrow files in python/R/stata
Plus a couple of drive by fixes for formatting and out of date stuff on the same page.

Fixes #1691 and https://github.com/opensafely-core/ehrql/issues/2244

### DIFF
--- a/docs/actions-scripts.md
+++ b/docs/actions-scripts.md
@@ -43,47 +43,70 @@ These types of outputs are considered potentially highly-disclosive, should not 
 Pseudonymised patient-level outputs tend to be large in size and therefore it is important that the right files formats are used for these large data files. The wrong formats can waste disk space, execution time, and server memory. The specific formats used vary with language ecosystem, but they should always be compressed.
 
 !!! note
-    The template sets up the `ehrql` command to produce `csv.gz` outputs.
-    This is the current recommended output format, as CSV files compress well,
-    and this reduces both storage requirements and improves job execution times
-    on the backend.
 
-    If you need to view the raw CSV data locally, you can unzip with `opensafely unzip dataset.csv.gz`.
+    The current recommended output format is `arrow` (Apache Arrow format). Arrow is a
+    language-independent columnar memory format which allows for efficient analytic operations
+    and supports compression.
 
+    However, for initial development, arrow formats are not easily inspected, so the template sets up the `ehrql` command to produce `csv.gz` outputs.
+
+    Both `csv.gz` and `arrow` format are compressed, which both reduces storage requirements and improves job execution times on the backend.
+
+    If you need to view raw CSV data locally, you can unzip with `opensafely unzip dataset.csv.gz`.
+
+    You are strongly recommended to use `arrow` format to run your analyses on the real data.
 
 
 === "Python"
 
     ```python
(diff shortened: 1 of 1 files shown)


docs/actions-scripts.md
```diff
@@ -43,47 +43,70 @@ These types of outputs are considered potentially highly-disclosive, should not
...
-    This is the current recommended output format, as CSV files compress well,
-    and this reduces both storage requirements and improves job execution times
-    on the backend.
 
-    If you need to view the raw CSV data locally, you can unzip with `opensafely unzip dataset.csv.gz`.
+    The current recommended output format is `arrow` (Apache Arrow format). Arrow is a
+    language-independent columnar memory format which allows for efficient analytic operations
+    and supports compression.
 
+    However, for initial development, arrow formats are not easily inspected, so the template sets up the `ehrql` command to produce `csv.gz` outputs.
```
evansd, line 51: Thinking about it now, `.csv.gz` isn't a great default because it's _neither_ the recommended production format _nor_ the easiest format to work with locally. I wonder if we should go one way or the other and either make the template use `.csv` or make it use `.arrow` and hope that the new VSCode tooling makes it easy enough for people to inspect the output of their datasets.

That's not a question for this PR though, just noting it will it occurs to me.

evansd (approved):

docs/actions-scripts.md
rebkwok: I did wonder about this, but I didn't think it was easy to view arrow files locally. Does VS Code have an extension that lets you view arrow files? Or are you referring to the `show()` function?

rebkwok: hmm, there are a couple in the VS marketplace, but nothing official. Maybe it's worth adding something to our extension

bloodearnest: So, I made the change to the template to generate csv.gz rather than csv a good while back.

There reason was that people were not changing it, and we ended up which huge uncompressed csv's in production, which a) consumed our limited disk indefinitely (as we don't delete) and b) make the prepare stage much much slower due to copying ~10x the amount of data.

It sucks, but I am keen not to go back to uncompressed by default.

evansd: Right, so that suggests maybe changing the default to `.arrow` is the right thing here.

Another nice consequence of switching our installation method to `uv` would be that this would work out-of-the-box as a way of viewing Arrow files:
```
uvx --with pyarrow visidata some_file.arrow
```

rebkwok: yes, definitely wasn't suggesting we go back to uncompressed csv as the default. I think the options are to leave the template as csv.gz or (preferably) change it to arrow, as long as people can easily view the output file locally

PR #1774, opened 2025-05-14
Rewrite case control docs to be compatible with ehrQL

alarthast:


docs/case-control-studies.md
```diff
@@ -114,58 +117,110 @@ actions:
...
+This allows us to only include the matched controls in our dataset.
+```python
+from ehrql import create_dataset
+
+dataset = create_dataset()
+dataset.define_population(
+    matched_patients.exists_for_patient()
 )
 ```
+Don't forget to add additional population constraints to the dataset if you require them!
```
alarthast, line 153: I included this sentence because someone has put additional constraints in a study - if we don't expect this to be a common occurrence, we can take out this sentence. 
https://github.com/opensafely/openprompt_health_utilisation/blob/1051dd9b84071aab6a6353d4801d8e6c17f3af63/analysis/dataset_definition_hx_matched_comp.py#L30-L32

docs/case-control-studies.md
```diff
@@ -1,7 +1,7 @@
 ## Background
 
 Data extracts often use an "index date" to define variables such as age, or presence of comorbidities at the start of a study.
-This can either be a fixed date (e.g. "2020-02-01") or a date that varies between patients, also known as [dynamic dates](legacy/study-def-dates.md#dynamic-dates) (e.g. the date that people were hospitalised).
```
alarthast, line 4: I have looked at the legacy docs linked here and I think they are more confusing than they are helpful now that cohort extractor has been deprecated, so I have removed the link.

I haven't replaced the link with anything else though, hopefully the concept of dynamic dates is understandable without needed an explanatory link?

docs/case-control-studies.md
```diff
@@ -36,64 +33,69 @@ actions:
...
 
-In this step, we will construct [a study definition](legacy/study-def.md) to extract all the data we need for the cases:
+In this step, we will construct [a dataset definition](../ehrql/tutorial/building-a-dataset) to extract all the data we need for the cases:
 that is, the *matching data*, or the data we will use to match the cases to the potential controls;
 and the *non-matching data*, or the data we will use for analysis.
+
 To avoid duplicating code to extract the matching data and the non-matching data in this and the following steps,
-we could use separate Python scripts to [share common study definition variables](legacy/study-def-tricks.md#sharing-common-study-definition-variables).
+we could use separate Python scripts to share common variables or parametrise the dataset definition.
+Your copilot would be happy to help you with this.
```
alarthast, line 42: This is a bit of a cop-out by [offering someone else's time](https://en.wiktionary.org/wiki/%E6%85%B7%E4%BB%96%E4%BA%BA%E4%B9%8B%E6%85%A8), but given that the legacy doc hadn't been replaced, sharing variables is a common practice and everyone gets a co-pilot now, hopefully this is okay

docs/case-control-studies.md
```diff
@@ -36,64 +33,69 @@ actions:
...
+
+As we will construct multiple dataset definitions in this and the following steps,
+we will name this dataset definition `dataset_definition_cases.py`.
 
-As we will construct multiple study definitions in this and the following steps, we will name this study definition *study_definition_cases.py*.
-When working with [multiple study definitions](legacy/study-def.md#multiple-study-definitions), each study definition's suffix is used to name each corresponding output file.
-Here, the study definition's suffix is *cases*, so the corresponding output file will be named *input_cases.csv.gz*.
+When working with multiple dataset definitions, it is good practice to use the same suffix to name each corresponding output file.
```
alarthast, line 47: Again, I have simply removed the link to the legacy docs here.

docs/case-control-studies.md
```diff
@@ -36,64 +33,69 @@ actions:
...
-Our *project.yaml* now includes the following action:
+Our `project.yaml` now includes the following action:
 
 ```yaml
 # ...
 actions:
   # ...
   matching:
-    run: python:v2 python analysis/match.py
+    run: python:latest python analysis/match.py
```
alarthast, line 98: Is this a superfluous change? Sorry if there is a reason for `python` to be at `v2`.

docs/case-control-studies.md
```diff
@@ -114,58 +116,104 @@ actions:
...
 
-In this step, we will construct a third study definition to extract only the non-matching data for the matched controls.
-We will name this study definition *study_definition_controls.py*.
+In this step, we will construct a third dataset definition to extract only the non-matching data for the matched controls.
+
+We will name this dataset definition `dataset_definition_controls.py`.
 
-We will use the functions `patients.which_exist_in_file` and `patients.with_value_from_file`.
-We will use `patients.which_exist_in_file` to include only the matched controls in the population:
+We will use the `@table_from_file` feature in ehrQL to make a table called `matched_patients` from the `matched_matches.csv.gz` file.
```
alarthast, line 123: I'll plan to ask nicely for someone from RAP to give the ehrQL code here a :eyes: .

docs/case-control-studies.md
```diff
@@ -114,58 +116,104 @@ actions:
...
+When we are done, `matched_patients` would behave as if it were any other
+[`PatientFrame`](../ehrql/reference/language/#PatientFrame)
+in ehrQL.
 
+Suppose `matched_matches.csv.gz` has the following columns: `patient_id`, `age`, `sex` and `case_index_date` -
+i.e. `age` and `sex` were the *matching data* we extracted in the first step.
```
alarthast, line 129: Hopefully I got the concept of "matching data" correct here?

docs/case-control-studies.md
```diff
@@ -114,58 +116,104 @@ actions:
...
+dataset = create_dataset()
+dataset.define_population(
+    matched_patients.exists_for_patient()
 )
 ```
+Don't forget to add additional population constraints to the dataset if you require them!
 
-We will use `patients.with_value_from_file` to access the case index dates:
+Since the `case_index_date` column in `matched_patients` is now accessible,
+we can use it to define columns of *non-matching data* in our dataset.
```
alarthast, line 157: Again, hopefully I got the concept of "non-matching data" correct here.

docs/case-control-studies.md
```diff
@@ -114,58 +116,104 @@ actions:
...
+For example, we might want to see if our matched controls have had a codelist event on or after their case index date.
+```python
+from ehrql import codelist_from_csv
+from ehrql.tables.core import clinical_events
 
+codelist = codelist_from_csv("codelists/codelist.csv")
+events_in_codelist = clinical_events.where(
+    clinical_events.snomedct_code.is_in(codelist)
+)
+dataset.has_event_in_codelist = events_in_codelist.where(
```
alarthast, line 167: Since this is a direct translation of the old docs, I've kept the `has_event_in_codelist` column as it is, rather than adding anything fancy. I've also not added any further columns. We can always add more examples if it'd be useful?

docs/case-control-studies.md
alexwalkerepi: Yes that's fine.

docs/case-control-studies.md
alexwalkerepi:
```suggestion
we can use it to define the `index_date` for controls (since they by definition don't have an index date).
```

docs/case-control-studies.md
alexwalkerepi:
```suggestion

```

docs/case-control-studies.md
alexwalkerepi:
```suggestion
    run: python:v2 python analysis/match.py
```

alexwalkerepi (changes requested):
I think this looks good as a direct translation from cohortextractor to ehrQL. A few comments/suggestions.

docs/case-control-studies.md
alarthast: Since the variable is still called `case_index_date`, how about:
`...we can use it as the index date for controls (since they by definition don't have an index date).`
?

docs/case-control-studies.md
alarthast: Instead of using RAP's time, I have instead verified that the ehrQL works by making a small case control study. See https://github.com/alarthast/case-control-example/pull/1.

docs/case-control-studies.md
alarthast: It turns out that `v2` is not right - it uses pandas' [deprecated `append` syntax](https://pandas.pydata.org/pandas-docs/version/1.4/reference/api/pandas.DataFrame.append.html), which errors out on the `v2` image.

So, we ought to keep `latest`, which is `v1`.

Error here: https://github.com/alarthast/case-control-example/actions/runs/15196524552/job/42741853267

iaindillingham (approved):
Thanks for picking this up, @alarthast. I think you've tested the code in the snippets by creating a toy repo, is this correct? Is the toy repo available? It will be useful to the next person who needs to update the documentation. Couple of comments:

* Is the warning about the TPP backend correct?
* There's a reference to `python:latest`. Should that be `python:v1`?

docs/case-control-studies.md
alarthast: At Iain's suggestion, we will be explicit and write `v1`. Captured this conversation in the message of a0b5a7407697aabb45bfbaf34a379228c7f07067.

PR #1799, opened 2025-06-26
Tweak mentions of cohort-extractor

madwort:


rebkwok (approved):
I wonder if we can just generate and commit the cohort-extractor legacy docs, since they'll never change now, but maybe we want to keep them as reference for old studies. Not having it as a git submodule would be nice when you're running the docs site locally - it's the thing that slowest to update

madwort:
> generate and commit the cohort-extractor legacy docs, since they'll never change now

that's a good idea - I've added it to this ticket:

https://github.com/opensafely/documentation/issues/1800

PR #1845, opened 2025-10-02
Skip `lycheeverse/lychee-action` 2.6.1.

mikerkelly:
2.6.1 seemed to cause the exclude-path to not take effect, waiting for next release.

Ran link checking action against this branch, passes. https://github.com/opensafely/documentation/actions/workflows/check_links.yml


StevenMaude (approved):
Thanks! We can see if a later version fixes our issue, or if we need to re-investigate.

PR #1858, opened 2025-10-20
Update container diagram to reflect RAP API v2 work

### LINKED ISSUE #1856: Update container diagram for RAP API V2 work
Update the [container diagram](https://docs.opensafely.org/technical-architecture/#container-diagram) to reflect the new direction of communication between job-server and the RAP controller.

rebkwok:
Basically just flips the arrow between job-server and the RAP controller.

Fixes #1856


madwort:
Good to reverse the arrow, although I'm not sure why it just says "fetches new jobs" when it is also sending the status updates (the old arrow mentioned both).

PR #1876, opened 2025-11-25
Document list of legally restricted codes

### LINKED ISSUE #978: Add more information on legally restricted codes
https://bennettoxford.slack.com/archives/C010SJ89SA3/p1663968080522169?thread_ts=1663926312.235539&cid=C010SJ89SA3

lucyb:
[Slack](https://bennettoxford.slack.com/archives/C03GWV1QL0K/p1764004885814149) :thread: 

Fixes #978

### DIFF
--- a/docs/data-sources/systmone.md
+++ b/docs/data-sources/systmone.md
@@ -40,6 +40,15 @@ These are coded in response to data flowing from Pillar 1 and Pillar 2. (At the 
 All coded events should be accompanied by an event date.
 However, invalid or unusual dates are possible and may occur for a number of reasons; for instance, when a GP is coding an important event in the patient's medical history but the exact date is not known.
 
+### Sensitive datasets
+
+Clinical codes relating to certain conditions or situations, as defined in the Royal College of GP’s (RCGP) sensitive datasets, are not available in OpenSAFELY. These are:
+
+* **Gender related issues**: SNOMED Refset for General Practice summary data sharing exclusion for gender related issues _999004351000000109_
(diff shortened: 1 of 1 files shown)


HelenCEBM:
Thanks Lucy! 

1. I think it's easier for readers to spot the key terms that they might need to know about if we start each point with a brief description in bold rather than the code e.g.:
> - **Gender related issues**: SNOMED Refset for General Practice summary data sharing exclusion for gender related issues 999004351000000109

2. Also it will be most helpful if we can provide links to the codes which are contained within each codelist. The browser links expire after a time I think. Can we create them all in OpenCodelists? I made one [here](https://www.opencodelists.org/codelist/user/helen-curtis/restricted-termination-codes/0d0deac0/)

lucyb:
> Just picking up a review of this. Technically the term 'legally restricted codes' refers to restrictions to certain ICD-10, OPCS and SNOMED-ECDS codes which should not be included in SUS datasets.
> 
> The codes which are excluded in our primary care data are excluded due to RCGP 'sensitive datasets' restrictions - link [here](https://digital.nhs.uk/services/summary-care-records-scr/viewing-guidance-including-additional-information). I think it might make sense to update the wording, and possibly include a link.

Thanks Rose. I've changed the title and wording, please let me know what you think?

I'm unsure about including the link, since it refers to the Summary Care Record and I'm concerned that it might be misleading. That said, if you're confident that it will be more meaningful to researchers (it's not an area I know that much about), then I can include a [direct link](https://digital.nhs.uk/services/summary-care-records-scr/viewing-guidance-including-additional-information#summary-care-record-exclusion-set) to the relevant section.

PR #1973, opened 2026-05-05
Update Job Server images

tomodwyer:
As they were using the old design.

- [Before](https://docs.opensafely.org/jobs-site/)
- [After](https://tomodwyer-job-server.opensafely-docs.pages.dev/jobs-site/)


StevenMaude (approved):
There's a slight nit that the unedited `project-edit-button.png` on the page has the current design, but uses an oval red ring to highlight a button.

But this is an improvement, so let's go with it, and we can always tidy up the other image later.

PR #2009, opened 2026-07-13
Update and standardise GitHub Actions

tomodwyer:
- Fix formatting with Prettier
- Bump Actions to latest version
- Fix auto-fixable issues with zizmor

### DIFF
--- a/.github/workflows/check_links.yml
+++ b/.github/workflows/check_links.yml
@@ -38,10 +38,18 @@ jobs:
       # This workaround may not be needed in future if lychee has better rate limiting.
       # https://github.com/lycheeverse/lychee/issues/36
       - name: Check links
-        # 2.6.1 seemed to cause the exclude-path to not take effect, waiting for next release.
         uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8 # v2.9.0
         with:
           args:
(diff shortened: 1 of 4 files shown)


.github/workflows/check_links.yml
```diff
@@ -39,8 +39,9 @@ jobs:
       # https://github.com/lycheeverse/lychee/issues/36
       - name: Check links
         # 2.6.1 seemed to cause the exclude-path to not take effect, waiting for next release.
-        uses: lycheeverse/lychee-action@82202e5e9c2f4ef1a55a3d02563e1cb6041e5332  # v2.4.1
+        uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8 # v2.9.0
```
StevenMaude, line 41: thought: This bumps the version of `lychee-action`. As in the comment above, I don't know if we figured out the issues with the update for `lychee`. Maybe the link check workflow needs testing, if it hasn't been. It can be run via manual workflow dispatch for the branch.

StevenMaude (approved):
Thanks! One thought, but looks fine besides.

.github/workflows/check_links.yml
tomodwyer: Nice, didn't realise I could manually run it, I'll test that now and either roll it back or remove the comment.

tomodwyer: Some arg updates required, but it looks to all be working now 🕺

PR #2013, opened 2026-07-23
Update text to recommend Python v3.10+

tomodwyer:
As we have updated OpenSAFELY CLI to drop support for Python v3.8 and v3.9

https://github.com/opensafely-core/opensafely-cli/issues/280

### DIFF
--- a/docs/install-python.md
+++ b/docs/install-python.md
@@ -4,7 +4,7 @@
 For security, consistency, and readability, OpenSAFELY provides an API built in [**Python**](https://www.python.org/) for using the platform.
 This API includes script-based functions for specifying the patients and variables that make up a study dataset (using [ehrQL](ehrql/index.md)),
 and command line functions for importing codelists, generating dummy data, and testing that the study definition can be run successfully on the server.
-**Python version 3.7 or higher** must be installed on your machine to perform these tasks.
+**Python version 3.10-3.13** must be installed on your machine to perform these tasks.
 
 Many functions are provided in a Python module called `opensafely` which will also need to be installed &mdash; see the [`opensafely` CLI section](opensafely-cli.md) for more details.
 


docs/index.md
```diff
@@ -37,7 +37,7 @@ To use OpenSAFELY, users must know, or be willing to learn, the following progra
 ### Essential
 
 - **Stata, R or Python**
-  OpenSAFELY currently supports Stata v16.1, Python 3.8, Python 3.10, R 4.0 and R 4.4 for statistical analysis.
+  OpenSAFELY currently supports Stata v16.1, Python 3.10 (and higher), R 4.0 and R 4.4 for statistical analysis.
```
mikerkelly: This is referring to the versions of Python available in the `python:v1` and `python:v2` images, not the CLI version. This has not changed and so this text should not change.

There is a `python:v3` coming, I think.

mikerkelly (commented):
Thanks Thomas.

The supported `opensafely-cli` versions are now 3.10, 3.11, 3.12, and 3.13. that does not include Python 3.14, [versions of which](https://www.python.org/downloads/release/python-3140/) have been released since 2025-10.

Given that, I think it would be more helpful to state explicitly that versions between 3.10--3.13 are supported. As if someone has the most recent Python, 3.14, as we recommend, they will not be able to use it right now. That could be frustrating, because we told them something not true, and depending on their control over their system the install may not have been a trivial thing to do (for example, if they had to seek special permission to have 3.14 installed, only to find that this does not work, and they have to go back and ask again).

What do you think?

I think that is relevant to `docs/install-python.md`, and also `README.md` from `opensafely-cli` which recommends "at least version 3.10 but the more recent the better". I think that we should change both of those.

I don't think we should change `docs/index.md`, see comment.

tomodwyer:
@mikerkelly I think I was being optimistic in hoping the Python v3.14 changes would be done before we would need to merge this, but you're right I'll add another PR in for that after this one.

docs/index.md
tomodwyer: Bah I ignored the Python images file assuming this one would be ok to change, but yes, in context it is wrong. There's probably a better way to phrase that sentence when v3 is out so we don't repeat ourselves in multiple files.

mikerkelly:
> @mikerkelly I think I was being optimistic in hoping the Python v3.14 changes would be done before we would need to merge this, but you're right I'll add another PR in for that after this one.

Fair. I think that any "version X+" when we actually require a bounded range is hostage to any future release making it inaccurate without us noticing. So it might be best to always state the exact range. Would be nice if we could somehow configure that once as a variable and have it propagate in the docs, probably worth it only if very quick though.

tomodwyer:
Good point, we don't automatically support 3.15 when it's out.

mikerkelly (approved):

Issue #2026, opened 2026-08-04
Remove GitHub Deployments


tomodwyer:
Our current `pages-deployment.yml` workflow includes a `gitHubToken`:

https://github.com/opensafely/documentation/blob/05fe727dedbdcb92787623ff4778a8a96e56aa75/.github/workflows/pages-deployment.yml#L46

This means that a new [GitHub Deployment is created for each build](https://github.com/opensafely/documentation/deployments):

<img width="1234" height="812" alt="" src="https://github.com/user-attachments/assets/8902c02e-9659-46b8-99c5-14ce4f5aa674" />

We do not use these deployments, as that is handled via Cloudflare. To stop this from happening, we can remove the token, as it is optional [as per the docs](https://github.com/cloudflare/wrangler-action/blob/84c72882ef1dacec6f70cdca9666d4625494e009/README.md#deploy-your-pages-site-production--preview).

We can also remove the existing deployments via the API, removing thousands of old deployment links.

## To do

- [x] [Remove Deployments from GitHub](https://github.com/opensafely/documentation/deployments)
- [x] [Remove Environments from GitHub](https://github.com/opensafely/documentation/settings/environments)
- [x] [Remove `gitHubToken` from `pages-deployment.yml`](https://github.com/opensafely/documentation/pull/2027)

PR #2027, opened 2026-08-04
Remove gitHubToken from deployment step

tomodwyer:
As per the Cloudflare Docs[^1]:

> Optional: Enable this if you want to have GitHub Deployments
> triggered

We do not use GitHub Deployments, so do not need this enabled

Part of #2026

[^1]: https://github.com/cloudflare/wrangler-action/blob/84c72882ef1dacec6f70cdca9666d4625494e009/README.md#deploy-your-pages-site-production--preview


mikerkelly (approved):
Thanks, Thomas.

Tiny feedback: it could be useful to put a link to the documentation site in the git commit message for the benefit of the reader.

I assume that was https://github.com/marketplace/actions/deploy-cloudflare-pages-github-action or equivalent. Presumably there are many places with Cloudflare documentation.

tomodwyer:
@mikerkelly good point, I had the link in the issue, but it's better here. I've added that in and rebased.

PR #2054, opened 2026-09-04
HES/SUS ICD-10 "X" padding

Jongmassey:
Document special ICD-10 format used in APCS dataset.


rose-higgins (approved):
Looks fine to me, thanks for updating

PR #2057, opened 2026-09-09
Update workflow instructions for dataset definition

HelenCEBM:
Removed lines for: specifying expected distributions for dummy data (now not really possible I think and anyway dummy data covered separately in step 3); and importing external files in the workflow.


rose-higgins (approved):
Looks great, thank you Helen

